PR #34, opened 2015-09-26
Added downloadFile sugar method

universome:
I think, this lib should has some sugar methods and downloading file by file_id is just musthave (I can't imagine a lot of other use cases for ```getFile``` api method.

BTW, [python's most popular telegram api lib](https://github.com/leandrotoledo/python-telegram-bot#api)  has such kind of functional

P.S. If you'd like to accept this PR, i think you should accept that PR first: https://github.com/yagop/node-telegram-bot-api/pull/33

~~P.P.S. Aligning has broken on several lines. I will fix it a little bit later, fixing it now is a pain for me, sry.~~ Fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,4 +264,19 @@ See: https://core.telegram.org/bots/api#sendlocation
 
 * **Promise**
 
+## getFile(fileId)
+
+Get file.
+Use this method to get basic info about a file and prepare it for downloading.
+Attention: link will be valid for 1 hour.
+
+See: https://core.telegram.org/bots/api#getfile
+
+### Params:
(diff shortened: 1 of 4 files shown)


universome:
Hm, my 3rd commit [(Added downloadFile sugar method)](https://github.com/universome/node-telegram-bot-api/commit/14dbd12f68caa25bd64f12d76e6e632c6a4f6295) is in master, but PR is not merged. How did it happen?

yagop:
I would swear I've written here days ago.
I merged by hand the issue, maybe that's why isn't closed the issue.

Thanks a lot @universome :smile:

PR #39, opened 2015-10-16
fix - serialize `reply_markup`

imposibrus:
@see https://core.telegram.org/bots/api#sendmessage


yagop:
Better check if it's an object and then serialize it. Notice that on exampled is passed serialized

yagop:
Done in 6a6111360b92f2b907822d13a4ba41a2aa79e64a

PR #41, opened 2015-10-18
allow 0ms for polling interval

### LINKED ISSUE #20: Use 0ms as polling interval value
Hi!

I would like to perform long polling like this:

```js
var bot = new TelegramBot(token, { polling: { timeout: 2000, interval: 0 } });
```

The problem is that 0 is replaced with default 2000ms inside the module. I can use `interval: 1`. It solves the issue but doesn't feel right. Is there any other way to perform long polling?

brandly:
fixes #20. thanks!!

by the way, i'm seeing a lot of tests fail locally due to `PEER_ID_INVALID`. it looks like i should probably set a `TEST_USER_ID`, but i'm not sure where to get that value from. thanks again

### DIFF
--- a/src/telegramPolling.js
+++ b/src/telegramPolling.js
@@ -17,7 +17,7 @@ var TelegramBotPolling = function (token, options, callback) {
   this.token = token;
   this.callback = callback;
   this.timeout = options.timeout || 0;
-  this.interval = options.interval || 2000;
+  this.interval = (typeof options.interval === 'number') ? options.interval : 2000;
   this.lastUpdate = 0;
   this.lastRequest = null;
   this.abort = false;


yagop:
I don't know why people want to do 0ms pooling interval, its really annoying for server and client, its just like DDoSing Telegram.

`TEST_USER_ID` is a user id (or chat) to send the messages. A bot can only send messages to a user if he has sent a message before.

brandly:
hahahah that's fair. i just saw the issue and figured i could fix it easily.

how's the CI server supposed to work, if keys aren't hardcoded? AppVeyor is throwing `'Bot token not provided'`

yagop:
For security Travis and AppVeyor don't expose environment variables on pull requests. I set up Travis to not building on PR, AppVeyor now doesn't build too.

brandly:
thanks for the explanation!

if there's no need for 0ms polling, you're more than welcome to close this PR. in that case, i'd close #20 as well

yagop:
You are welcome.
Done, thanks.

PR #42, opened 2015-10-18
add nodejs 4.x to travis

brandly:


yagop:
Sorry I don't see any points to add nodejs 4.x to Travis. The more versions are added, the more time takes and more API calls are done (may lead on a client ban). Maybe one could be added but if its working on 0.12.x it should be working on 4.x

PR #43, opened 2015-10-19
Updated Example to use onText

reyy:


yagop:
Thanks!

In stead of `msg.from.id`, `msg.chat.id` must be used.

reyy:
Whoops fixed! Thanks!

PR #46, opened 2015-11-09
Add filename parsing

rkhitin:
Parsing filename from remote files stream let show right name in message. For example 'cat.gif' instead 'undefined', and telegram show preview for img in case if we read stream by url.

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -98,6 +98,13 @@ TelegramBot.prototype._request = function (path, options) {
   options = options || {};
   options.url = this._buildURL(path);
   debug('HTTP request: %j', options);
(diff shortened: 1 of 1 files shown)


yagop:
It shouldn't be implemented on `_request` funciton but in `formatSendData`

rkhitin:
Now it's in `formatSendData` and little bit cleaner.

GochoMugo:
Is it possible to add an example, where this is necessary?

GochoMugo:
This PR will be closed after some time! Please provide more information!

GochoMugo:
Closing this Cold PR! It got stale!

PR #50, opened 2015-11-23
Automatically JSON.stringify keyboard markup

christophetd:
I noticed that you currently have to manually call JSON.stringify on your keyboard markup when you need one.
This automatically calls JSON.stringify on the keyboard_markup option if it specified and was not already stringified


yagop:
Thanks bit was previosly done

6a6111360b92f2b907822d13a4ba41a2aa79e64a

christophetd:
Indeed, I didn't see it was in _sendRequest. Maybe it would be a good idea to update the example then?

Issue #45, opened 2015-10-26
Message error 


kirillrybin:
With 1st sample from readme.md
![Error om message](http://i.imgur.com/1S4yzKK.png)

yagop:
Take a look on the example, when a message arrives, it will sends an image called `cats.png`
```
// Any kind of message
bot.on('message', function (msg) {
  var chatId = msg.chat.id;
  // photo can be: a file path, a stream or a Telegram file_id
  var photo = 'cats.png';
  bot.sendPhoto(chatId, photo, {caption: 'Lovely kittens'});
});
```
If it doesn't exits it will fail

GochoMugo:
This has been fixed by replacing it with a simpler example (commit 75d940c473c6ec252ee24ba4a788ecf458227a68)!

Thanks @kirillrybin for your interest in this library! :+1:

PR #54, opened 2016-01-01
Fixes the readme with an appropriate example that doesn't crash

AndrewRayCode:
Addresses #45 

Please consider the experience of newcomers and the user experience design rule of [reasonable defaults](http://tutorials.jenkov.com/api-design/provide-sensible-defaults.html) when creating documentaiton :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,13 @@ bot.onText(/\/echo (.+)/, function (msg, match) {
   bot.sendMessage(fromId, resp);
 });
 
-// Any kind of message
+// Log the message and echo user's text
 bot.on('message', function (msg) {
-  var chatId = msg.chat.id;
-  // photo can be: a file path, a stream or a Telegram file_id
-  var photo = 'cats.png';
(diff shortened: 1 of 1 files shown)


yagop:
Hi, thanks for your feedback. You are pointing to "sensible defaults" and I can't see anything related with documentation. 

I think it would be cool to provide functions on the callback to response the messages without sending the `chatId` but its out of scope of this module. I don't know if that is what you mean with "reasonable defaults".

On README.md I provide an overview of the module. There are examples on the [`examples`](https://github.com/yagop/node-telegram-bot-api/tree/master/examples) folder too. I thought if someone reads `cats.png` will assume he needs a `cats.png` file, but maybe I were wrong (I assume many things).

I will consider it, thanks.

PS: Happy new year.

GochoMugo:
I think this PR is appropriate as it allows anyone run the samples, without requiring more information on the whys and hows. This might reduce cases where it results in issues being created in the issue-tracker.

Another alternative would be adding more comments to the samples.

GochoMugo:
Closing this as we are addressing this in PR #224.

PR #70, opened 2016-01-30
Added `waitResponse` utility method

rlogiacco:
I believe this method will be very appreciated as it greatly simplifies user interaction into bots!

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -45,6 +45,7 @@ var TelegramBot = function (token, options) {
   ]; // Telegram message events
   this.textRegexpCallbacks = [];
   this.waiting = {}; // Waiting for response
+  this.options.response = this.options.response || {timeout: 20000};
   this.processUpdate = this._processUpdate.bind(this);
 
   if (options.polling) {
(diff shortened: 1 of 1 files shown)


yagop:
Really interesting I like it too much!. I was working with another solution which don't uses promises (but a callback that can be executed many times).

yagop:
Closed by 17cb062c9642313f17c3c5e070de2032031857d3

rlogiacco:
I see your implementation is quite different.... I personally prefer promises over callbacks... Do you mind if I publish an alternative npm module?

yagop:
I prefer promises over callbacks too but a Promise can't be called (resolved) more than once. There can be multiple replies to the same message.

yagop:
Of course, you can alternative npm module, this is OpenSource :blue_heart:

rlogiacco:
> There can be multiple replies to the same message.
I believe we are taking two different approaches, but I understand your point.

I've just added another feature, an `onCommand` function which splits arguments by commas: a little less flexible than `onText` but simpler for the end user in the most common cases, I believe.

Have a look at my fork if you wish, I will not submit a pull request

:+1:

PR #74, opened 2016-02-04
Added `removeReplyListener` as the counterpart for `onReplyToMessage`

githugger:
Adds the function `removeReplyListener` to `TelegramBot`, which removes a previously registered callback (created using `onReplyToMessage`).
This way the developer can make sure his application does not retain unnecessary callback functions in the `onReplyToMessages` array in his `TelegramBot` instance.
Also, if so desired, this can be used to easily ensure, that only on response to a message will be processed, as shown in the following short code snippet:

```
// bot is an instance of TelegramBot
bot.sendMessage(someChatId, someMessage, someOpts)
  .then(function(sentMsg) {
    bot.onReplyToMessage(sentMsg.chat.id, sentMsg.message_id, function(reply) {
      if (isValidReply(reply) {
        // do something ... and:
        removeReplyListener(sentMsg.chat.id, sentMsg.message_id);
      }
    });
  });
```

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -107,7 +107,7 @@ TelegramBot.prototype._processUpdate = function (update) {
           // Responding to that message
           if (reply.messageId === message.reply_to_message.message_id) {
             // Resolve the promise
-            reply.callback(reply.id);
+            reply.callback(message);
           }
         }
       });
@@ -573,17 +573,19 @@ TelegramBot.prototype.onText = function (regexp, callback) {
  * Register a reply to wait for a message response.
  * @param  {Number|String}   chatId       The chat id where the message cames from.
  * @param  {Number|String}   messageId    The message id to be replied.
- * @param  {Function} callback     Callback will be called with the reply 
+ * @param  {Function} callback      Callback will be called with the reply 
+ * @return {Number} newId           The id of the onReplyToMessages object.
  * message.
  */
 TelegramBot.prototype.onReplyToMessage = function (chatId, messageId, callback) {
-  var newId = newReplyId();
+  var newId = this.newReplyId();
   this.onReplyToMessages.push({
     id: newId,
     chatId: chatId,
     messageId: messageId,
     callback: callback
   });
+  return newId;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


yagop:
I think its better if `onReplyToMessage` returns a id and `removeReplyListener` takes the id as argument.

githugger:
I know what you mean, it feels a bit more clean your way.

I actually implemented the remove and onReply messages with ids for the reply objects in the `onReplyToMessages` array now.

The id's are created via a simple index counting up, but that gives 16 billion id's, that's 12 billion leeway for the maximum `onReplyToMessages` array-size, so I hope that will be alright.

rithvikvibhu:
Any progress on this?

src/telegram.js
```diff
@@ -567,15 +573,42 @@ TelegramBot.prototype.onText = function (regexp, callback) {
...
+    var reply = this.onReplyToMessages[i];
+    if (reply.id === replyId) {
+      index = i;
+      break;
+    }
+  }
+  if (index === -1) {
+    return null;
+  }
+  var removedListener = this.onReplyToMessages.splice(index, 1);
```
GochoMugo: > An array containing the deleted elements. If only one element is removed, an array of one element is returned. If no elements are removed, an empty array is returned.
>
> From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

Return the **first and only element** of the returned array!

For example,

```js
var removedListeners = this.onReplyToMessages.splice(index, 1);
return removedListeners[0];
```

src/telegram.js
```diff
@@ -54,6 +54,12 @@ var TelegramBot = function (token, options) {
   if (options.webHook) {
     this._WebHook = new TelegramBotWebHook(token, options.webHook, this.processUpdate);
   }
+
+  // id generator for reply messages
+  var idx=0;
+  this.newReplyId = function() {
```
GochoMugo: This naming may bring name-conflicts in the future.

This would be preferable:

```js
let replyListenersIndex = 0;
this._nextReplyListenerId = function() { return replyListenersIndex++; };
```

We need the `_` in the name to signify it is **not** exported and should **not** be used by library consumers.

GochoMugo:
PR merged in commit [79de62a96e98e8516d2388da473d7a6a33f9d9b4](https://github.com/yagop/node-telegram-bot-api/commit/79de62a96e98e8516d2388da473d7a6a33f9d9b4).

Thanks @githugger for your time and effort in this PR!

GochoMugo:
Shipped in **v0.27.0**.

PR #80, opened 2016-03-01
URL-decoded filename when sending file from stream

RafaelKr:
This sends the file as "My File.ext" instead of "My%20File.ext".

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -277,7 +277,7 @@ TelegramBot.prototype._formatSendData = function (type, data) {
     formData[type] = {
       value: data,
       options: {
-        filename: fileName,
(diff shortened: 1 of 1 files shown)


yagop:
How about using [`querystring.unescape`](https://nodejs.org/api/querystring.html#querystring_querystring_unescape)?

```js
var qs = require('querystring');
...
filename: qs.unescape(fileName)
```

RafaelKr:
I didn't know about that function, thank you. I've updated the code.

yagop:
Thank you! :smiley:

RafaelKr:
Awesome! My first Pull Request, thank you too! 😄

PR #90, opened 2016-03-17
refactor: rewrite to babel, es6 & eslint

AVVS:
I don't really know how strong is the need to support 0.10.x and 0.12.x, but I've refactored the code into es6 along with some babel for unsupported stuff.

summary of changes:

1. native classes
2. bound functions instead of doing that in the constructor
3. request-promise instead of promisify over request
4. pump wrapping streams, so that handlers dont leak in case of errors
5. bluebird 3
6. removed should from tests, used native assert + is, overall tests seem to still be sucky because they only validate type of the returned data. I'd say they need more work with .reflect() of the promise and more strict checks
7. jshint -> eslint because, well, eslint wins with es6
8. bufferlist for body parsing on the webhook server - since it's JSON, you don't want to convert it to utf-8 and then JSON.parse - that works just fine on a Buffer and is faster because you don't do 2 conversions
9. style remained in tact
10. moved all parts with try/catch to separate functions to avoid making those unoptimizable - there are 3 instances of those in each class, so maybe you would want to have a single function in a separate file
11. short hand properties instead of `options || {}`

I understand that this PR might not be desirable, but, in case you actually fill otherwise, or plan to upgrade it some day - I did the grunt work :)


idchlife:
This is actually interesting.

yagop:
I started the repository back in 2015 where ES6 wasn't even popular. I were thinking to rewrite it to ES6, thanks @AVVS for the work!

I will take a closer look and merge it soon.

I would rather use XO but I don't mind ESLint as default in the proyect.

Kudos!

yagop:
I've made some changes. @AVVS what do you think?

knock-in:
> JSDoc can't parse class-properties, generate lib-doc.

Hi @yagop,
I made a pull request with some small changes to fit the JSDoc parser, so you don't need `lib-doc` anymore. See #94

PR #91, opened 2016-03-17
Update README.md

EXL:
Fix typo


yagop:
Thanks! README.md is generated with https://github.com/yagop/node-telegram-bot-api/blob/master/README.hbs

PR #93, opened 2016-03-19
Make Linter Happy

nikhiljha:
Make the bithound linter happy by adding use strict statements/single quotes for strings.


GochoMugo:
Bithound is so happy right now! We made many changes recently! See https://www.bithound.io/github/yagop/node-telegram-bot-api/master/files

Thanks @nikhiljha for your time and effort!

PR #94, opened 2016-03-21
Generate Docs from /src directly

knock-in:


yagop:
Merged. Dunno why isn't marked for GitHub

Issue #116, opened 2016-05-02
If use switch between webHook or Polling you get an 409 (http://telegram.wiki/bots:boterrorcode) telegram error.


w-site:
If use switch between webHook or Polling you get an 409 (http://telegram.wiki/bots:boterrorcode) telegram error.

w-site:
https://github.com/yagop/node-telegram-bot-api/pull/117

w-site:
@yagop Hi. What do you think about this issue ?

PR #117, opened 2016-05-02
Fix: If use switch between webHook or Polling you get an 409 (http://telegram.wiki/bots:boterrorcode) telegram error. #116

w-site:
Fix: If use switch between webHook or Polling you get an 409 (http://telegram.wiki/bots:boterrorcode) telegram error.


AVVS:
@yagop please merge this in - its quite helpful

yagop:
Published in `0.23.2`.
Time to go to bed for me hehe.

PR #121, opened 2016-05-06
add the sendVenue method

iiroj:
This pull request adds the method and test for sendVenue. This method was added to the Telegram Bot API on April 9, 2016 in the Bot API 2.0 update. See https://core.telegram.org/bots/api#sendvenue for more info.

I did this blindly in GitHub because the change seems simple enough. I did briefly test it live in my Telegram bot and it sent the venue correctly.

I attached a screenshot of how the venue looks in the iOS client:
img_0885

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -490,6 +490,28 @@ class TelegramBot extends EventEmitter {
     form.longitude = longitude;
     return this._request('sendLocation', { form });
   }
+  
+  /**
+   * Send venue.
+   * Use this method to send information about a venue.
+   *
+   * @param  {Number|String} chatId  Unique identifier for the message recipient
+   * @param  {Float} latitude Latitude of location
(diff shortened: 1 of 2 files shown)


iiroj:
Oh, by the way, there is a new optional key _foursquare_id_ that only works with _sendVenue_. I'm not sure if I should have done anything to it, so I just left it out, assuming it'll just work if added in the additional query options.

icdevin:
Looks good -- and you're right about the last bit. Most things that are no explicitly set by the methods are implicitly shoved into the query options and passed along all the same.

OrKoN:
Hey, when will this be merged? seems useful to me

GochoMugo:
Thanks @iiroj for your time and effort!

yagop:
I gave @GochoMugo collaborator rights, thanks for your work guys!

PR #129, opened 2016-05-21
fix sendLocation for action parameter + add sendVenue method

antonrifco:
fix sendLocation for action parameter + add sendVenue method

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -448,7 +448,7 @@ class TelegramBot extends EventEmitter {
    */
   sendChatAction(chatId, action) {
     const form = {
-      action,
+      action: action,
       chat_id: chatId
     };
     return this._request('sendChatAction', { form });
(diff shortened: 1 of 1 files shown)


src/telegram.js
```diff
@@ -448,7 +448,7 @@ class TelegramBot extends EventEmitter {
    */
   sendChatAction(chatId, action) {
     const form = {
-      action,
+      action: action,
```
w-site: why? in ES6 you did't do this

src/telegram.js
```diff
@@ -490,6 +490,28 @@ class TelegramBot extends EventEmitter {
...
+   * @return {Promise}
+   * @see https://core.telegram.org/bots/api#sendlocation
+   */
+  sendVenue(chatId, latitude, longitude, title, address, form = {}) {
+    form.chat_id = chatId;
+    form.latitude = latitude;
+    form.longitude = longitude;
+    form.title = title;
+    form.address = address;
+    return this._request('sendVenue', { form });
```
w-site: Why you need form object if you send directly values that you provide in argument of function ???
just
` return this._request('sendVenue', { 
chatId,
latitude,
longitude,
title,
address
 });`
keep it simple =)

w-site:
can you add test for new function ?

src/telegram.js
maddin77: Because the form argument is used for [optional parameters](https://core.telegram.org/bots/api#sendvenue) (`reply_to_message_id`, `reply_markup`, etc.).

Also 
```js
return this._request('sendVenue', { chatId, latitude, longitude, title, address });
```
wont work since request requires form data to be passed via `form` property.

w-site: @maddin77 Thanks for answer =)

src/telegram.js
GochoMugo: This is valid ES6 syntax. Please see https://github.com/lukehoban/es6features#enhanced-object-literals for more information on **shorthand assignments**.

src/telegram.js
GochoMugo: This has been added in PR #121.

GochoMugo:
We are closing this PR due to the reasons stated in the @GochoMugo's review.

Thank you @antonrifco for your time and effort.

Issue #109, opened 2016-04-21
new-inline-keyboards Bot 2.0


nullifye:
does this module support new [inline keyboard](https://core.telegram.org/bots/2-0-intro#new-inline-keyboards) right now?

icdevin:
You can set inline buttons with static URLs but you can't respond to callback_query POSTs. Watch #102.

edlvj:
@icdevin Can you add code examples of using InlineKeyboardMarkup and InlineKeyboardButton?

icdevin:
Sure thing. Note that the inline keyboard property is an array of arrays where each inner array is a row of buttons, so the below example gives 3 row-width buttons. You can make one nested array with 3 objects inside to make a single row with 3 buttons inside it.

```
var options = {
  parse_mode: 'Markdown',
  disable_web_page_preview: true,
  reply_markup: JSON.stringify({
    inline_keyboard: [
      [{ text: 'Some button text 1', callback_data: '1' }],
      [{ text: 'Some button text 2', callback_data: '2' }],
      [{ text: 'Some button text 3', callback_data: '3' }]
    ]
  })
};
bot.sendMessage(msg.chat.id, 'Some text giving three inline buttons', options).then(function (sended) {
  // `sended` is the sent message.
});
```

Scram99:
Cool, that is exactly what i was looking for. But if I understand you correctly this lib "node-telegram-bot-api" does not support the "bot.on('callback_query)" at the moment!? :-( 
That's such a pity. Will that be implemented one day? Unfortunately I'm not skilled enough to do this.

edlvj:
@icdevin thnx for reply

icdevin:
I forked @tjhorner and added a few lines to support this on my branch [here](https://github.com/icdevin/node-telegram-bot-api/tree/callback-query). You can use that but it's still missing some (one?) group-level feature and tests. Otherwise just watch #102.

nullifye:
how to use your fork @icdevin? ive already cloned you fork but cannot start.

>telegram.js:17
>class TelegramBot extends EventEmitter {
>^^^^^
>SyntaxError: Unexpected reserved word
>    at Module._compile (module.js:439:25)
>    at Object.Module._extensions..js (module.js:474:10)
>    at Module.load (module.js:356:32)
>    at Function.Module._load (module.js:312:12)
>    at Module.require (module.js:364:17)
>    at require (module.js:380:17)
[...]

PR #102, opened 2016-04-12
Update for Telegram Bots 2.0

tjhorner:
# Telegram Bots 2.0!

I've added the new methods from the updated Bots API in Telegram. This includes all of the methods for kicking/banning users as well.

The APIs for kicking and unbanning chat members don't seem to work right now because it says the bots must be admins in the groups, but there's no way to make bots admins in groups quite yet. (unless I'm missing something...)

EDIT: It seems that the developers are aware that users cannot make bots admins and are fixing it. Personally, I think they're trying a slow rollout.

### Methods added

- [`kickChatMember`](https://core.telegram.org/bots/api#kickchatmember)
- [`unbanChatMember`](https://core.telegram.org/bots/api#unbanchatmember)
- [`answerCallbackQuery`](https://core.telegram.org/bots/api#answercallbackquery)
- [`editMessageText`](https://core.telegram.org/bots/api#editmessagetext)
- [`editMessageCaption`](https://core.telegram.org/bots/api#editmessagecaption)
- [`editMessageReplyMarkup`](https://core.telegram.org/bots/api#editmessagereplymarkup)

PR #132, opened 2016-05-23
Adding callback_query

tionsys:
Support for responses of Inline Buttons in messages or Inline Queries

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -611,6 +611,48 @@ TelegramBot.prototype.sendLocation = function (chatId, latitude, longitude, opti
 };
 
 /**
+ * Send venue.
+ * Use this method to send information about a venue.
+ *
+ * @param  {Number|String} chatId  Unique identifier for the message recipient
+ * @param  {Float} latitude Latitude of location
+ * @param  {Float} longitude Longitude of location
+ * @param  {String} title Name of the venue
+ * @param  {String} address Address of the venue
+ * @param  {Object} [options] Additional Telegram query options
+ * @return {Promise}
+ * @see https://core.telegram.org/bots/api#sendvenue
+ */
+TelegramBot.prototype.sendVenue = function (chatId, latitude, longitude, title, address, options) {
+  var form = options || {};
+  form.chat_id = chatId;
+  form.latitude = latitude;
+  form.longitude = longitude;
+  form.title = title;
+  form.address = address;
+  return this._request('sendVenue', {form: form});
+}
+
+/**
+ * Send contact.
+ * Use this method to send phone contacts.
+ *
+ * @param  {Number|String} chatId  Unique identifier for the message recipient
+ * @param  {String} phone_number Contact's phone number
+ * @param  {String} first_name Contact's first name
+ * @param  {Object} [options] Additional Telegram query options
+ * @return {Promise}
+ * @see https://core.telegram.org/bots/api#sendcontact
+ */
+TelegramBot.prototype.sendContact = function(chatId, phoneNumber, firstName, options) {
+  var form = options || {};
+  form.chat_id = chatId;
+  form.phone_number = phoneNumber;
+  form.first_name = firstName;
+  return this._request('sendContact', {form: form});
+}
+
+/**
  * Get file.
  * Use this method to get basic info about a file and prepare it for downloading.
  * Attention: link will be valid for 1 hour.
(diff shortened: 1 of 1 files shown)


highfeed:
+1

Kaos1337:
Can you kindly provide an example as requested here https://github.com/yagop/node-telegram-bot-api/issues/109#issuecomment-221112685?

tionsys:
Hey,
data is returned via callback_query like the Telegram Bot Api Manual states.
"msg" is an instance of CallbackQuery (https://core.telegram.org/bots/api#callbackquery)
```js
var api = new TelegramBotApi("your-token", {polling: true});

api.on('callback_query', function(msg) {
	var user = msg.from.id;
	var data = msg.data;
	api.sendMessage(msg.from.id, "You clicked button with data '"+ data +"'");
});

var options = {
	reply_markup: JSON.stringify({
	inline_keyboard: [
		[{text: 'Some button text 1', callback_data: '1'}], // Clicking will send "1"
		[{text: 'Some button text 2', callback_data: '2'}], // Clicking will send "2"
		[{text: 'Some button text 3', callback_data: '3'}]  // Clicking will send "3"
    ]
  })
};

api.sendMessage(msg.from.id, "Click a button to display data", options);
```

Kaos1337:
Do you have any suggestion if I would:
- send a message with the inline_keyboard after the execution of a command (/mycommand)
- wait for the user interaction
- continue the execution of the code inside the command /mycommand

Why I'm asking that, well, I'd like to make a bot that uses the inline_keyboard interaction to make 2 or 3 questions to the user and at the end of all elaborate them.
An example:
- user: /calculate
- bot: choose one of these [ 1, 2, 3]
- user: <make a choise>
- bot: choose one of these [ 4, 5, 6]
- user: <make a choise>
- bot: the result is: choise1+choise2

tionsys:
You could write a wrapper around the sendMessage funtion returning a promise or similar.
The CallbackQuery has a field called "message" where you can read the chat and message id from. 

Example:
```js
// Example wrapper
var listeners = [];
function sendButtonsMessage(text) {
	return new Promise(function(resolve, reject) {
		api.sendMessage(to, text, buttons).then(function(message) {
			listeners[listeners.length] = {chat: message.chat.id, id: message.message_id, listener: resove}
		});
	});
}

// Fetch a matching listener
api.on('callback_data', function(data) {
	for(var i = 0; i < listeners.length; i++) {
		if(listeners[i].chat != data.message.chat.id) continue;
		if(listeners[i].id != data.message.message_id) continue;
		listeners[i].listener(data);
		// Uncomment this if you want every listener to only become executed one time
		// listeners.splice(i, 1);
	}
});

// Example
sendButtonsMessage("hello").then(function userClickedAButton(button) {
	api.sendMessage(to, "You clicked on button "+ button.data);
});
```

I honestly don't recomment doing it this way, because you might have thousands of users at some time and you don't want to store that much callbacks.

Kaos1337:
I decided to use an sql server to record every message sent that contains inline buttons with a "state", when I receive a callback_data event I check the message id and I know in which step the user is, so I can edit the inline buttons to continue the procedure and update the "state" into the db.

paOol:
@kaos1337 can you post some pseudo code of your implementation? 

Also, couldn't you a state management library like redux instead of requiring a database

Kaos1337:
I'm using a database because it's something that I've already adopted for that bot, so, the addition of another table wasn't a big challenge. That table records all steps where a request should go when received. (yes, it's like a final states machine)
I have a class that manages:
- the INSERT of an inline request when I've to create a new process to elaborate.
- the UPDATE of an inline request when I've change the next step.
- the DELETE of an inline request when the process finish.

The api callback_data event function has just to compare (SELECT) the message id (of the request itself) with the next step (it's all into the db) and call the relative function that allows the prosecution of the desired process. You have to keep in mind that all inline buttons could be called at any point of the time... obviously I'll deny all request older than some hours. You'll enjoy the editTextMessage call a lot.
That's all.

I'll give a look at the suggested lib.

yagop:
`callback_query` added in #102

rodrigosalinas:
Hi. Can you please provide a very simple example of how to use answerCallbackQuery? I don't know how to use it. Thanks in advance.

Kaos1337:
@rodrigosalinas just call it like sendMessage: bot.answerCallbackQuery(params...);

PR #133, opened 2016-05-25
JSON-stringify reply_markup when it is sent in qs

m-burst:
Currently, `reply_markup` is only JSON-stringified when sent in form (e.g. in `sendMessage`). However, `sendPhoto` tries to send it in query string. I've fixed this.


GochoMugo:
Thank you @m-burst for your time and effort! :+1:

PR #144, opened 2016-06-10
getChat method

serhiidmytruk:
Hi! I added getChat method — https://core.telegram.org/bots/api#getchat


GochoMugo:
Thanks @serhiidmytruk for your time and effort in this PR! :+1:

PR #147, opened 2016-06-11
use indexOf to match url

AVVS:


src/telegramWebHook.js
```diff
@@ -64,7 +64,7 @@ class TelegramBotWebHook {
     debug('WebHook request headers: %j', req.headers);
 
     // If there isn't token on URL
-    if (!this.regex.test(req.url)) {
+    if (!this.urlValidator(req.url)) {
```
GochoMugo: Why don't you call `indexOf()` here, so that it be:

```js
if (req.url.indexOf(this.token) === -1) {
```

Seems simpler and avoids the func call (although the code may be automatically optimized for that, by v8).

GochoMugo:
Done in commit [9d12bdfa41e5f22062543a33cd53798e61dc489f](https://github.com/yagop/node-telegram-bot-api/commit/9d12bdfa41e5f22062543a33cd53798e61dc489f).

Thanks @AVVS!

GochoMugo:
Shipped in **v0.27.0**.

PR #150, opened 2016-06-14
Provided example with inline keyboard and callbacks

agseco:
I missed an example like this when I tried this library, so I have added it for other developers.


GochoMugo:
This PR inspired the [commit 0174b875ff69f6750fc80a049315c9a0d7a5e471](https://github.com/yagop/node-telegram-bot-api/commit/0174b875ff69f6750fc80a049315c9a0d7a5e471). Thanks @agseco!

PR #169, opened 2016-07-13
Fixed a couple of debug prints

oflisback:


GochoMugo:
Thank you @oflisback for your effort!

PR #178, opened 2016-07-31
Proxy extra request options

tarmolov:
I have a server which has only ipv6 access to api.telegram.org. So, I must pass `family: 6` option for each request to telegram api.

It would be great to add an extra option to TelegramBot constructor which makes it possible.

@yagop what do you think?


GochoMugo:
This PR is awesome! I'm ready to merge this! Please rebase your branch!

However, the polling requests are **not** using these custom options.

yagop:
Please don't use extend, use`Object.assign`.

PR #191, opened 2016-09-15
Add edited_message event

chris54721:
Adds support for the `edited_message` parameter in Updates.
Emits an `edited_message` event when a message is edited, and also two specific events, `edited_message_text` and `edited_message_caption`, depending on the type of the message.

### DIFF
--- a/README.hbs
+++ b/README.hbs
@@ -32,8 +32,8 @@ bot.on('message', function (msg) {
 There are some other examples on [examples](https://github.com/yagop/node-telegram-bot-api/tree/master/examples).
 
 ### Events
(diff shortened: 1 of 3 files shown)


GochoMugo:
These events seems quite useful IMO. Thanks for taking your time to work on this.

kamikazechaser:
Awesome...I was looking for this feature!

seufry:
I tried it, but I do not get the `inline_query` or `chosen_inline_result` events when I push button 'Play GAME_NAME'. This is my mistake?

chris54721:
This PR is unrelated to inline queries, it adds support for message editing. The "Play GAME_NAME" button from the gaming platform is probably not considered an inline query.

seufry:
@chris54721 oh, I wrote at the wrong PR :(

GochoMugo:
@yagop What do you think about this PR? I am okay with merging it in.

yagop:
Perfect but I dont know if `reply_markup` can be edited too.

GochoMugo:
It seems `reply_markup` can be edited too. See [editMessageReplyMarkup](https://core.telegram.org/bots/api#editmessagereplymarkup).

yagop:
I pointed it out because currently it is missing. 😝

chris54721:
That's not an event though, since bots can't receive messages from other bots

GochoMugo:
@chris54721 is right. Also a [Message](https://core.telegram.org/bots/api#message) does **not** have a `reply_markup` property!

GochoMugo:
I have merged in this PR. Thanks @chris54721 for your time and effort. :+1:

Issue #77, opened 2016-02-14
Performance improvement


preco21:
I'm not pretty sure about this. But when I enable [keep-alive](https://en.wikipedia.org/wiki/HTTP_persistent_connection) via [`forever` option of `request` module](https://github.com/request/request#requestoptions-callback), its response speed was much better.

This is simple `console.time` benchmarks:

```
// with keep-alive
t: 503.079ms 
t: 496.879ms 
t: 496.488ms 
t: 494.888ms 
t: 496.991ms 
t: 495.026ms 
t: 496.277ms 
t: 498.538ms 
t: 511.290ms 

// without keep-alive
t: 1116.574ms
t: 1013.224ms
t: 1123.053ms
t: 1045.034ms
t: 1047.880ms
t: 1087.082ms
t: 1058.906ms
t: 1080.411ms
t: 1056.534ms
```

Cause since this module used `request`, monkey patch used (ES6):

``` javascript
import {Request} from 'request';

const initBase = Request.prototype.init;

Request.prototype.init = function(options = {}) {
  options.forever = true;
  return Reflect.apply(initBase, this, [options]);
};
```

How about it?

yagop:
Thanks! I will chek it out.

preco21:
It seems on `master` branch it has already been fixed. 👍 
See: https://github.com/yagop/node-telegram-bot-api/blob/master/src/telegramPolling.js#L105

preco21:
Sorry, repoen because of `_request()` method still does not have `forever` option: https://github.com/yagop/node-telegram-bot-api/blob/master/src/telegram.js#L145

preco21:
Alternative PR: https://github.com/yagop/node-telegram-bot-api/pull/178

PR #193, opened 2016-09-26
Fix request performance issue

preco21:
Fixes: https://github.com/yagop/node-telegram-bot-api/issues/77

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -157,6 +157,7 @@ class TelegramBot extends EventEmitter {
     options.url = this._buildURL(_path);
     options.simple = false;
     options.resolveWithFullResponse = true;
+    options.forever = true;
     debug('HTTP request: %j', options);
     return request(options)
       .then(resp => {


GochoMugo:
I am **not** sure you have demonstrated that performance is improved by using the [forever-agent](https://github.com/request/forever-agent).

GochoMugo:
But we are already using the `forever` option during polling. I do **not** see why we shouldn't use the option while doing other requests against the API.
This PR seems fine to me.

/cc @yagop

preco21:
@GochoMugo I have been using my forked version of module for my personal bot. It is working properly with better performance. I think since polling uses `forever` option already, It would be nice to have same option for requests against the APIs. 👍

GochoMugo:
@yagop, merge this?

GochoMugo:
PR #178 will allow one to **proxy more custom options** for the requests.

So, instead of applying the `forever` option for everyone as the default, we could instead **add note on the `forever` option in our Readme** and let users choose whether to use it or not. This helps **maintain backwards-compatibility**.

preco21:
@GochoMugo I agree about the idea. I can write some note about this feature.

GochoMugo:
@preco21 Nice! That would allow us to keep our API stable. Should we decide to break it, we might add this option.

preco21:
@GochoMugo Since polling uses `forever` option already, you can consider about adding option that directly adding `forever` option to requst functions without through proxy object for taking balance of the API.

GochoMugo:
Thanks @preco21 for your time and effort!

preco21:
@GochoMugo Awesome! ✨

Issue #202, opened 2016-10-04
Support of request cancellation


dimotsai:
I was working on my project, trying to cancel some API requests as following:

``` js
const TelegramBot = require('node-telegram-bot-api');

let telegramBot = new TelegramBot('BOT_TOKEN' , {polling: false})

let doAsync = () => telegramBot.sendPhoto(CHAT_ID, 'wallpaper.jpg');

doAsync().cancel();
```

It seems like that `doAsync()` is not cancellable. The request won't terminate until the photo is sent.
## 

I found the current dependency is `request-promise@v2.0.1`. That will be nice if you could upgrade `request-promise` to the latest version!

A part of changelog from `request-promise`:

```
v4.1.0 (2016-07-30)
* Added cancellation support (Thanks to @not-an-aardvark for pull request #123)
```

GochoMugo:
This has been added in v0.25.0!

Thanks @dimotsai for your interest in this library!

dimotsai:
@GochoMugo Good job! Thanks for supporting. This is a nice library :)

Issue #52, opened 2015-12-26
Possible memory leak?


bjesus:
Hi!
I'm running two bots on the smallest droplet on DigitalOcean (512MB), and I'm hitting 100% memory usage almost constantly. I noticed that even just running the following code causes the node process to start consuming more and more memory:

```
var TelegramBot = require('node-telegram-bot-api');
var nconf = require('nconf');
nconf.argv().env().file({ file: 'config.json' });
var token = nconf.get('bot_token');
var bot = new TelegramBot(token, {polling: true});
```

After initializing `bot` in the fifth line - the process's memory usage starts to grow and never stops.

I'm not sure how to debug it further. Any ideas?

I tried limiting node with `--max_old_space_size` but it doesn't seem to work (same behavior). Both bots are just rarely used, and anyhow it seems like the leak starts before my own code is even reached.

crisbal:
I am interested in this too. I also notice a small but constant memory usage increase.

yagop:
The leak might be on https://github.com/yagop/node-telegram-bot-api/blob/master/src/telegramPolling.js#L50 there is an infinite loop which I should fix ...

crisbal:
Could this be fixed by replacing the timeout with a set interval?

yagop:
Could be if its called outside. The matter of using an interval is that the function could be called before the previously one haven't finished. That behavior must be avoided.

yagop:
> I tried limiting node with --max_old_space_size but it doesn't seem to work (same behavior). Both bots are just rarely used, and anyhow it seems like the leak starts before my own code is even reached.

Seems the option is `--max-old-space-size`.

Which NodeJS version are you using? Seems v4 have improved in memory usage. I changed the `_polling` method a bit, seems the memory still increasing in node v0.12.4 but slowly and hope it converge to smaller number. With v4.2.4 increases till 50MB and stops.

bjesus:
I tried both `--max-old-space-size` and `--max_old_space_size` before, none seemed to work.

Anyhow, I just upgraded Node.JS's version to 5.4 (was running 0.10.25...) and upgraded node-telegram-bot-api, and the problem seem to be gone. it's been 24 hours I can't recognize no signal for the leak.

Thank you!

GochoMugo:
With the recent changes, I believe this issue might be fixed. We are also running bots using this library, and we do **not** have any memleak problems.

[...]

PR #211, opened 2016-10-19
fixed #207 

### LINKED ISSUE #207: request-promise causes memory leaks
The current version of request-promise looks like contain a memory leak (related to #52) because of internal cache implementation according to Chrome profile. I've updated `package.json` with request-promise ^4.1.1 and it looks the problem with leaks was mitigated. 

You can reproduce the problem by increasing timer interval to 10 and leave it for an hour.

desunit:
updated request-promise package to eliminate memory leaks

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "mime": "^1.3.4",
     "pump": "^1.0.1",
     "request": "^2.69.0",
-    "request-promise": "^2.0.1"
+    "request-promise": "^4.1.1"
   },
   "devDependencies": {
     "babel-cli": "^6.6.5",


GochoMugo:
Just for formality, have you ensured we have satisfied all the migration guidelines, [as stated in the `request-promise` documentation](https://github.com/request/request-promise#readme)?

desunit:
The major changes related to transform which is not used in this code. We upgraded the code for our production and it works for more than week without memory leaks and issues.

CapacitorSet:
Very nice. I have been using this "in production" and the server no longer crashes.

desunit:
@CapacitorSet, glad to hear that!

GochoMugo:
My summary:

* [x] ~~... make sure that request is installed into your project as a direct dependency ...~~ we already installing request as a direct dependency!
* [x] ~~... Continuation Local Storage is no longer supported ...~~ we do **not** use Local Storage!
* [x] ~~... your transform function ...~~ we do **not** use Transform function!
* [x] ~~... handling of the transform function got overhauled ...~~ we do **not** use Transform function!
* [x] ~~... Bluebird got updated from v2 to v3 ...~~ we already using v3

We are ok to merge this! Will be in next release!

Thanks @desunit for your time and effort!

GochoMugo:
This closes issues #202 and #207.

desunit:
Thanks @GochoMugo! I think it should close #52 too.

GochoMugo:
We might need to confirm with the original thread creator and contributors.

Issue #208, opened 2016-10-17
Implement getGameHighScores


yagop:
Implement method `getGameHighScores`.

https://core.telegram.org/bots/api#getgamehighscores

chrisonntag:
Hi @yagop, I'd like to work on this if you don't mind 😊

yagop:
![https://media.giphy.com/media/l3E6tfijzciPpgCdi/giphy.gif](https://media.giphy.com/media/l3E6tfijzciPpgCdi/giphy.gif)

chrisonntag:
I didn't find any tests for 'games' in general. Should i open an own issue on that? I'd like to test it first.

GochoMugo:
@chrisonntag I think it would be better to have a separate issue (or PR :smile:) for the tests. We really need help in our tests.

GochoMugo:
This has been added in v0.25.0!

Issue #203, opened 2016-10-05
When games api will be added?


highfeed:
When games api will be added?

GochoMugo:
PR #201 is addressing this. Watch it...

GochoMugo:
Some of the Game API methods have been added. Specifically, [TelegramBot#sendGame()](https://github.com/yagop/node-telegram-bot-api/blob/58566d2df5146611a6a3e6140ca616403d24a20e/README.md#TelegramBot+sendGame) and [TelegramBot#setGameScore()](https://github.com/yagop/node-telegram-bot-api/blob/58566d2df5146611a6a3e6140ca616403d24a20e/README.md#TelegramBot+setGameScore). Thanks to @jishnu7!

Also, @jishnu7 [promised us](https://github.com/yagop/node-telegram-bot-api/pull/201#issuecomment-252622527) to implement the remaining methods. :smile:

Will keep this issue open until the remaining methods are implemented and published.

GochoMugo:
This has been added in v0.25.0!

Thanks @highfeed for your interest in this library!

PR #213, opened 2016-10-22
adding getGameHighScores support

jishnu7:


GochoMugo:
This looks perfect to me. Completes the Games implementation in this library. Fixes issues #208 and #203.

Thanks @jishnu7 for keeping your promise. :+1:

Only thing missing now is its accompanying tests. :cry:

PR #216, opened 2016-10-25
Add option to ignore further regexes after match is found

GingerPlusPlus:
By default,  all regexes registered via `onText` are tried and in case of match callbacks are called, even if match is already found, which can cause one message to be processed multiple times.

I added `onlyFirstMatch` option which, as the name suggests, causes the bot to ignore further regexes after match is found.

It also enables people to add `.onText(/(?:)/, ...)` or something like that to handle the messages which were not matched by any regex.

I did not change existing behavior, you need to opt-in.

That's my first pull request, so I appreciate any hints you could give me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Emits `message` when a message arrives.
 | [options.webHook] | <code>Boolean</code> &#124; <code>Object</code> | <code>false</code> | Set true to enable WebHook or set options |
 | [options.webHook.key] | <code>String</code> |  | PEM private key to webHook server. |
 | [options.webHook.cert] | <code>String</code> |  | PEM certificate (public) to webHook server. |
+| [options.onlyFirstMatch] | <code>Boolean</code> | <code>false</code> | false: try matching all regexps; true: stop after first match |
 
 <a name="TelegramBot+stopPolling"></a>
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -537,7 +538,7 @@ username of a user, group or channel, etc.).
 
 | Param | Type | Description |
 | --- | --- | --- |
-| chatId | <code>Number</code> &#124; <code>String</code> | Unique identifier for the target chat or username of the target supergroup or channel (in the format @channelusername) |
+| chatId | <code>Number</code> &#124; <code>String</code> | Unique identifier for the target chat or username of the target supergroup or channel |
```
GochoMugo: I do not know why we needed to change this. :open_mouth:

src/telegram.js
```diff
@@ -41,6 +41,7 @@ class TelegramBot extends EventEmitter {
    * @param {Boolean|Object} [options.webHook=false] Set true to enable WebHook or set options
    * @param {String} [options.webHook.key] PEM private key to webHook server.
    * @param {String} [options.webHook.cert] PEM certificate (public) to webHook server.
+   * @param {Boolean} [options.onlyFirstMatch=false] false: try matching all regexps; true: stop after first match
```
GochoMugo: A better description would be sth along the lines of "If true, stop after first match. Otherwise, all regexps are executed".

README.md
GingerPlusPlus: Me neither. It just happened when I "regenerated" readme, I decided to leave it as-is. I guess somebody before me forgot to regenerate readme after this minor change.

src/telegram.js
GingerPlusPlus: Your wording is fine, feel free to reword; it's my first pull request, so I don't know how to change it.

I purposefully put documenting in separate commit than implementation, so one could be easily changed without other, because I wasn't sure if my wording was the best.

PR #224, opened 2016-11-11
Add simple example in README (PR #54)

GochoMugo:
Closes #54!

### DIFF
--- a/README.hbs
+++ b/README.hbs
@@ -9,22 +9,31 @@ npm install node-telegram-bot-api
 ```js
 var TelegramBot = require('node-telegram-bot-api');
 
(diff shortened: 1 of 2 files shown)


GochoMugo:
This should close issue #45.

By common logic, by the time a programmer is looking to send a photo, using `TelegramBot#sendPhoto()`, they would almost certainly be familiar with streams, etc. Such a programmer would be considered *advanced* to another who would only be interested in **sending text messages**, using `TelegramBot#sendMessage()`. This latter kind of programmer could be considered a *beginner* and yet it can also be an *advanced* one! We are targeting the **beginner**! *Advanced* programmers will understand the dcumentation!

PR #227, opened 2016-11-21
New Update Fields From API v2.3

kamikazechaser:
bot will now get updates about posts in channels. Added new fields channel_post and edited_channel_post to Update.

[v2.3 Update](https://core.telegram.org/bots/api/#update)

Example:

```js
bot.on('channel_post', function (msg) {
  var chatId = msg.chat.id;
  console.log(msg.text)
  bot.sendMessage(chatId, 'New Channel Post Detected');
});

bot.on('edited_channel_post', function (msg) {
  var chatId = msg.chat.id;
  bot.sendMessage(chatId, 'Channel Post Has Been Edited');
});
```

PR #239, opened 2016-12-16
add "channel_post" and "edited_channel_post" updates

backmeupplz:
As per https://core.telegram.org/bots/api-changelog#november-21-2016


GochoMugo:
Sorry, but this has been addressed in PR #227. However, thanks for your time and effort! :+1:

PR #245, opened 2016-12-21
Ensure consistency of methods signatures

GochoMugo:
### Bug:

The library assumes signatures of methods to be, somewhat:

```
methodName(requiredParam1, requiredParam2, form = {})
```

where `requiredParam1` (`requiredParam2`, ..., `requiredParamN`)
are parameters that MUST be provided, and
`form` is an optional object allowing supplying any additional,
optional parameters that the Bot API allows.

This allows any new parameters added by Telegram to be
readily-supported by our library.

### Notes:

This fix is **backwards-compatible**; while we should
notify the community of the change in method signatures,
old code using the old method signatures will **not**
break.

In the future, when we decide to release a major release
(i.e. bump major version), we could remove the code
providing this backwards-compatibility.


GochoMugo:
I also suggest:

1. Documenting the old signatures (for completeness)
1. Deprecation notice to allow users to migrate to new API with time

Should I proceed to include the above additions?

PR #252, opened 2017-01-04
Replaced bound functions with functions in prototype

GingerPlusPlus:
That enables the library to be used without transpiling on Node 6.9 (current LTS).

I'm pretty sure that the only difference in behavior is that when you do `const foo = webHookInstance._requestListener`, it currently gives you bound function, and after this change, it gives you unbound one; that shouldn't break anything I believe, but I haven't tested.

### DIFF
--- a/src/telegramWebHook.js
+++ b/src/telegramWebHook.js
@@ -45,7 +45,7 @@ class TelegramBotWebHook {
   }
 
   // pipe+parse body
-  _parseBody = (err, body) => {
+  _parseBody(err, body) {
     if (err) {
       return debug(err);
     }
@@ -58,8 +58,7 @@ class TelegramBotWebHook {
     return null;
   }
 
-  // bound req listener
-  _requestListener = (req, res) => {
+  _requestListener(req, res) {
     debug('WebHook request URL: %s', req.url);
     debug('WebHook request headers: %j', req.headers);
 


GochoMugo:
There are two functions, we need to bind:  (probably at L13)

```js
this._requestListener = this._requestListener.bind(this);
this._parseBody = this._parseBody.bind(this);
```

yagop:
Or something like this:

```js
this._webServer = https.createServer(opts, this._requestListener.bind(this));

 .pipe(bl(this._parseBody.bind(this)))
```

GingerPlusPlus:
Or, we could leave them bound, but move their definition to `constructor`.

GingerPlusPlus:
`parseBody` is only called from `_requestListener`, so it's not an issue; when we resolve `_requestListener`, `_parseBody` will be called with correct `this` as well.

GochoMugo:
@yagop, we need a definitive answer, so we can merge this in!

yagop:
Currently can't be merced. Functions must be bound.

GochoMugo:
This PR has been merged in [commit 487b5f4b8b0f839b6caea0318a1a3e0ae18877a0](https://github.com/yagop/node-telegram-bot-api/commit/487b5f4b8b0f839b6caea0318a1a3e0ae18877a0).

I have resolved to using:

```js
this._requestListener = this._requestListener.bind(this);
this._parseBody = this._parseBody.bind(this);
```

**Why?** The functions are bound **only once**.

Otherwise,

```js
.pipe(bl(this._parseBody.bind(this)))
```

will involving binding on each request (as it happens in the request listener!)

And,

```js
this._webServer = https.createServer(opts, this._requestListener.bind(this));
```

is cumbersome, as we would have to perform the binding for both the `http` and `https` cases!

Issue #36, opened 2015-10-03
The Offset Infinite Loop


vhdmsm:
Hi
My bot has been stuck in infinite loop and it's not responding anymore, what's problem?
here is the stack trace:

```
Unhandled rejection Error: 403 {"ok":false,"error_code":403,"description":"Error: Bot was kicked from a chat"}
    at /Users/vahid/Desktop/node_modules/node-telegram-bot-api/src/telegram.js:72:15
    at tryCatcher (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:577:18)
    at Promise._settlePromises (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:693:14)
    at Async._drainQueue (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
Unhandled rejection Error: 403 {"ok":false,"error_code":403,"description":"Error: Bot was kicked from a chat"}
    at /Users/vahid/Desktop/node_modules/node-telegram-bot-api/src/telegram.js:72:15
    at tryCatcher (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/vahid/Desktop/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:577:18)
[...]

PR #265, opened 2017-01-25
src/polling: Fix the Offset Infinite Loop bug

GochoMugo:
References:

  * Issue #36: https://github.com/yagop/node-telegram-bot-api/issues/36
    * answer/solution: https://github.com/yagop/node-telegram-bot-api/issues/36#issuecomment-268532067

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Changed:
 
 Fixed:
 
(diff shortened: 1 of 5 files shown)


GochoMugo:
While this bug fix may **not** able to be shipped in the next release, I would highly appreciate if I got some feedback on the implementation.

/cc @yagop @kamikazechaser @GingerPlusPlus @phillipadsmith

PR #272, opened 2017-01-29
Webhook health check endpoint

mironov:
Docker-ized environments usually expose an HTTP endpoint for health checks.

For example, in order to use a bot built with this lib in a kubernetes cluster, we have to give it an HTTP endpoint that responds 200 OK.

Refs:
https://github.com/kubernetes/contrib/tree/master/ingress/controllers/gce#health-checks
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#defining-a-liveness-http-request

### DIFF
--- a/src/telegramWebHook.js
+++ b/src/telegramWebHook.js
@@ -13,6 +13,7 @@ class TelegramBotWebHook {
    * @param  {String} token Telegram API token
    * @param  {Boolean|Object} options WebHook options
    * @param  {Number} [options.port=8443] Port to bind to
+   * @param  {String} [options.healthEndpoint=/healthz] An endpoint for health checks that always responds with 200 OK
    * @param  {Function} callback Function for process a new update
    */
   constructor(token, options, callback) {
@@ -27,6 +28,7 @@ class TelegramBotWebHook {
     this.options.https = options.https || {};
     this.callback = callback;
     this._regex = new RegExp(this.token);
+    this._healthRegex = new RegExp(options.healthEndpoint || '/healthz');
     this._webServer = null;
     this._open = false;
     this._requestListener = this._requestListener.bind(this);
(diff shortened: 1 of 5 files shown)


src/telegramWebHook.js
```diff
@@ -133,8 +133,13 @@ class TelegramBotWebHook {
     debug('WebHook request URL: %s', req.url);
     debug('WebHook request headers: %j', req.headers);
 
-    // If there isn't token on URL
-    if (!this._regex.test(req.url)) {
+    if (/\/healthz/.test(req.url)) {
```
GochoMugo: Could we make this configurable, like with `this._regexp`?

src/telegramWebHook.js
```diff
@@ -133,8 +133,13 @@ class TelegramBotWebHook {
...
+      // If this is a helth check
```
GochoMugo: uuuuuh! Typo! :smile_cat:

test/utils.js
```diff
@@ -31,6 +31,17 @@ exports = module.exports = {
...
+   * Send a message to the webhook at the specified port and path.
+   * @param  {Number} port
+   * @param  {String} path
+   * @param  {Object} [options]
+   * @param  {String} [options.method=POST] Method to use
+   * @param  {Object} [options.message] Message to send. Default to a generic text message
+   * @param  {Boolean} [options.https=false] Use https
+   * @return {Promise}
+   */
+  sendWebHook,
```
GochoMugo: What if we named it `sendWebHookRequest`, so it makes more sense?

mironov: Good call, fixed!

src/telegramWebHook.js
mironov: Ooops, thanks! 🙈

src/telegramWebHook.js
mironov: Sure! Done.

src/telegramWebHook.js
```diff
@@ -133,8 +135,13 @@ class TelegramBotWebHook {
...
     debug('WebHook request headers: %j', req.headers);
 
-    // If there isn't token on URL
-    if (!this._regex.test(req.url)) {
+    if (this._healthRegex.test(req.url)) {
+      // If this is a health check
+      debug('WebHook health check passed');
+      res.statusCode = 200;
+      res.end('OK');
+    } else if (!this._regex.test(req.url)) {
```
GochoMugo: I think we should test for the token **first**, then the health endpoint, considering that we do **not** want to waste too much CPU with each and every update we receive on the webhook! What do you think?

mironov: Makes sense, changed the order.

GochoMugo:
Thanks @mironov for your time and effort in this PR! Awesome! :tada:

GochoMugo:
Shipped in **v0.27.0**.

Issue #240, opened 2016-12-16
Way to multiple SendMessage in correct sequence


sidelux:
There is a way to send multiple Message in rapid sequence without change their order?

For example
```
bot.SendMessage(chat_id, "1");
bot.SendMessage(chat_id, "2");
bot.SendMessage(chat_id, "3");
```

Sometimes print
321 or 312 or 132

But i need
123

Thanks a lot!

yagop:
Haven't tested it, but something like:

```js
bot.SendMessage(chat_id, "1")
  .then(() => {
    return bot.SendMessage(chat_id, "2")
  })
  .then(() => {
    return bot.SendMessage(chat_id, "3")
  })
```

El vie., 16 dic. 2016 9:27, Edoardo Cortese <notifications@github.com>
escribió:

There is a way to send multiple Message in rapid sequence without change
their order?

For example

bot.SendMessage(chat_id, "1");
bot.SendMessage(chat_id, "2");
bot.SendMessage(chat_id, "3");

Sometimes print
321 or 312 or 132

But i need
123

Thanks a lot!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/yagop/node-telegram-bot-api/issues/240>, or mute the
thread

sidelux:
It works!

![image](https://cloud.githubusercontent.com/assets/18572941/21256782/9827467a-c374-11e6-91e3-c76a454c3a7e.png)

PR #305, opened 2017-03-21
src: add method to send messages sequentially

TheBeastOfCaerbannog:
- [x] All tests pass
- [x] I have run `npm run gen-doc`

### Description
Add a method which allows to send messages one-by-one. Promise returned by this method resolves to array of responses for each message. Sequence is also correct. It should correctly work for all .send* methods. 
I will add more tests to this PR if current implementation looks good enough.

### References
Issue #240

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -59,6 +59,7 @@ TelegramBot
     * [.sendGame(chatId, gameShortName, [options])](#TelegramBot+sendGame) ⇒ <code>Promise</code>
     * [.setGameScore(userId, score, [options])](#TelegramBot+setGameScore) ⇒ <code>Promise</code>
     * [.getGameHighScores(userId, [options])](#TelegramBot+getGameHighScores) ⇒ <code>Promise</code>
+    * [.sendSequence(method, chatId, messages, [options], [fileOpts])](#TelegramBot+sendSequence) ⇒ <code>Promise</code>
 
 <a name="new_TelegramBot_new"></a>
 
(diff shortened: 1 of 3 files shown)


GochoMugo:
First, thanks @TheBeastOfCaerbannog for a wonderful and clean PR! :100: 

As much as I would want to merge this in, I wish not to. This PR focuses on a common design/usage pattern by bot developers, while this library focuses on low-level interaction with the Bot API. Am afraid adding such code may add more technical debt and encourage more of such patterns to be added into the library.

Have you looked at other higher-level libraries, such as [tgfancy](https://github.com/GochoMugo/tgfancy), that focus on such patterns? I believe this would fit perfectly there!

Also, note that the above is my technical opinion on this matter, and is **not** the final decision on it. Am willing to listen to others' opinions. :smile:

/cc @yagop

yagop:
First of all, thank you so much for the PR and your time elaborating it.

As @GochoMugo says, I also think its out of the scope of this module the `sendSequence` method because the aim of the module is to keep only TG methods.

TheBeastOfCaerbannog:
Ok, thank you, I'll close this PR and will look at [tgfancy](https://github.com/GochoMugo/tgfancy) :)
But I guess [help.md](https://github.com/yagop/node-telegram-bot-api/blob/master/doc/help.md) is a bit misleading, because it says:
"How do I send multiple message in correct sequence?

Not done. PRs welcome!"

Issue #242, opened 2016-12-18
Add Middleware layer


yagop:
A middleware layer wold be very useful and not very hard to add. Express design is a nice example.

https://github.com/yagop/node-telegram-bot-api/issues/238#issuecomment-267839398

CapacitorSet:
Is there still interest? I could do this.

zandzpider:
yes there is !

kamikazechaser:
Implemented in an [external library](https://github.com/idchlife/node-telegram-bot-api-middleware) that adds onto ntba.

PR #313, opened 2017-04-25
Middleware

GuidoNebiolo:
- [ ] All tests pass
- [x] I have run `npm run gen-doc`

### Description

TelegramBot object now expose the method use to insert a middleware to manage the incoming message.

### References

Issue #242

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ output.md
 output/
 lib/
 lib-doc/
+.idea/
--- a/doc/api.md
+++ b/doc/api.md
@@ -96,9 +96,9 @@ Emits `message` when a message arrives.
 <a name="TelegramBot+use"></a>
 
 ### telegramBot.use(fn) ⇒ <code>this</code>
-Register a middleware function to be executed for every incoming message.
+Register a middleware function to be executed for every update.
 You can register more than one middleware, but you should be careful because is important the order.
-The middleware function has two parameters: the first one is the incoming message and the second one is
+The middleware function has two parameters: the first one is the incoming update and the second one is
 the next function, that should be called to continue the execution of the middleware chain.
 
 **Kind**: instance method of <code>[TelegramBot](#TelegramBot)</code>  
--- a/doc/api.md
+++ b/doc/api.md
@@ -12,6 +12,7 @@ TelegramBot
 
 * [TelegramBot](#TelegramBot)
     * [new TelegramBot(token, [options])](#new_TelegramBot_new)
+    * [.use(fn)](#TelegramBot+use) ⇒ <code>this</code>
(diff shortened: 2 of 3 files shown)


GochoMugo:
Is it possible to come up with an example illustrating a simple use case where this new feature is being used? It would really help in designing and validating the feature.

Otherwise, we are happy to see a PR! :100:

GuidoNebiolo:
Sure! for example it's possible to creating a simple authentication middleware like this one:
`
bot.use((message, next) => {
  const chatId = message.chat.id;
  if (chatId === 1234) {
    next();
  } else {
    bot.sendMessage(chatId, 'Chat unauthorized');
  }
});
`
or a middle for analytics like the issue #242 

It's possible to build a chain of many middleware and break the chain avoiding the next method calling.

GochoMugo:
Why are text messages (and **not** other types of messages) used in handling the middlewares? Is it possible to allow all types of messages?

GuidoNebiolo:
I can change the code and pass the update object (the same of the method processUpdate (https://core.telegram.org/bots/api#update) instead of message object inside the middleware.
Do you thing this could be useful?

GuidoNebiolo:
Now the middleware function accept as first argument the update object.
This is an example of how to use it:
```javascript
bot.use((update, next) => {
  console.log(update);
  next();
}).use((update, next) => {
  if (!!update.message) {
    const chatId = update.message.chat.id;
    if (chatId === 1234) {
      next();
    } else {
      bot.sendMessage(chatId, 'Chat unauthorized');
    }
  }
});
```
You can also chain more use method.

GochoMugo:
Things are looking better. Here are some suggestions:

1. It would be better if the `processUpdate()` method remain largely same as before. Just before emitting the relevant events, we would handle the middlewares (using a `_executeMiddlewares()` method, for example). 

2. How do the middlewares and the rest of update processing work together? Are events *always* emitted even if a middleware stops the chain execution i.e. events and middlewares are independent of each other. Or would it work like below, where the middlewares affect the events:

```js
processUpdate(...) {
    ...
    // the callback is invoked if all the middlewares invoke `next()` in their executions!
    return this._executeMiddlewares(update, function() {
        // the events are now emitted, etc.
        // ...
    });
}
```

What do you think?

MCSH:
IMO the events should be emitted only if all of the middlewares have called next. 

Also, would it be possible to change the event inside a middleware? An example would be that I have a "nickname" field in database for each userid and I'd like the 'chat' or 'from' part of each message to have that nickname as well, will I be able to do this?

GochoMugo:
Yeah, that should be possible, just like in Express. The middlewares can mutate the Update object.

noamcore:
This pull request was abandoned?

muety:
Would love to have support for middlewares!

Issue #325, opened 2017-05-09
onText global regex only works every second time


Sirius-A:
> Bug Report

I have read:
* [Usage information](https://github.com/yagop/node-telegram-bot-api/tree/master/doc/usage.md)
* [Help information](https://github.com/yagop/node-telegram-bot-api/tree/master/doc/help.md)

I am using the latest version of the library.

### Expected Behavior
When sending a _/get_ command, the bot replies everytime.

### Actual Behavior
The bot only answers every second time.

### Steps to reproduce the Behavior
1. Register a global regex with capture groups like so: 
`tgBot.onText(/\/get(Today)?(@\w+)?$/gmi,getTodayHandler);`
2. Send _/get_ to bot. (ok)
3. Send _/get_ to bot again. (No reply)

### Reason this happens
From http://stackoverflow.com/questions/11477415/why-does-javascripts-regex-exec-not-always-return-the-same-value:

> A JavaScript RegExp object is stateful.
> When the regex is global, if you call a method the same regex object, it will start from the index past the end of the last match.
> When no more matches are found, the index is reset to 0 automatically.

### Proposed solution
Reset the `lastIndex ` after the regexp.exec() call [here](https://github.com/yagop/node-telegram-bot-api/blob/9a9dfa9560eaf0bddb732868809be4c0e21585c4/src/telegram.js#L472).

I will put a PR together later today 😄

GochoMugo:
We definitely do **not** want the state of the RegExp object to be carried forward between different text messages. Otherwise, strange errors, that are difficult to reproduce, would occur occasionally.  We should reset the state.

Thanks @Sirius-A for this nice catch. Waiting for your PR. :+1:

Sirius-A:
> Otherwise, strange errors, that are difficult to reproduce, would occur occasionally.
That's exactly how I got here 💩

GochoMugo:
Sorry about that :rofl:.  Lemme check out the PR.

GochoMugo:
**New Release, v0.28.0 :confetti_ball:**

beshur:
Reproduces for me on 0.52.0

PR #326, opened 2017-05-09
fix(onText): reset state of global regexp after search

Sirius-A:
- [x] All tests pass
- [x] I have run `npm run gen-doc` (no changes)

### Description
Regex matching in `onText`: Resets the search index to start each regex match at the beginning again . 

**Note:** I usually add test cases to code I add/modif, but in this case it seems rather difficult to do that without modifying the `onText `function itself. I want to count the number of times the callback is executed. Any Ideas?

### References
Issue #325

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -473,6 +473,8 @@ class TelegramBot extends EventEmitter {
           if (!result) {
             return false;
           }
+          // reset index so we start at the beginning of the regex each time
+          reg.regexp.lastIndex = 0;
           debug('Matches %s', reg.regexp);
           reg.callback(message, result);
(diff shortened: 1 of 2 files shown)


GochoMugo:
How about something like this:

```js
const regexp = /ab*/g;

bot.onText(regexp, function() {
    // the 'lastIndex' property should be reset by now!
    assert.equal(regexp.lastIndex, 0);
});
```

Sirius-A:
:+1: Good idea. Thanks! 

I added the test to my commit.

GochoMugo:
Thanks @Sirius-A for this awesome PR! :+1:

PR #330, opened 2017-05-18
[API] v3 SendVideoNote method

kamikazechaser:
## API v3 

**sendVideoNote**

[[source](https://core.telegram.org/bots/api#sendvideonote)]

_There seems to be an issue with Telegrams documentation regarding this method, which doesn't state that length is actually a required parameter._

- [x] Add tests.
- [x] Generate documentation.
- [x] Run tests.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -33,6 +33,7 @@ TelegramBot
     * [.sendDocument(chatId, doc, [options], [fileOpts])](#TelegramBot+sendDocument) ⇒ <code>Promise</code>
     * [.sendSticker(chatId, sticker, [options])](#TelegramBot+sendSticker) ⇒ <code>Promise</code>
     * [.sendVideo(chatId, video, [options])](#TelegramBot+sendVideo) ⇒ <code>Promise</code>
+    * [.sendVideoNote(chatId, videoNote, length, [options])](#TelegramBot+sendVideoNote) ⇒ <code>Promise</code>
     * [.sendVoice(chatId, voice, [options])](#TelegramBot+sendVoice) ⇒ <code>Promise</code>
     * [.sendChatAction(chatId, action)](#TelegramBot+sendChatAction) ⇒ <code>Promise</code>
     * [.kickChatMember(chatId, userId)](#TelegramBot+kickChatMember) ⇒ <code>Promise</code>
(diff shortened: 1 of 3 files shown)


kamikazechaser:
PR is ready.

GochoMugo:
I would suggest separate PRs for the different methods. We could dedicate this one for the `sendVideoNote` method.

GochoMugo:
> There seems to be an issue with Telegrams documentation regarding this method, which doesn't state that length is actually a required parameter.

Have you contacted Bot Support?

kamikazechaser:
Yup, waiting for their reply.

kamikazechaser:
> Hey there, you're quite right, it's currently mandatory, which should be fixed soonish. For any new note however it will be mandatory, existing notes by file_id will be optional. For now, please supply it though.

GochoMugo:
@kamikazechaser We can go ahead and classify it as an *optional* parameter. It should go in the `options` object. We should follow the **correct** API but we should add a **note/disclaimer** for the users that the parameter may be *required* until the API is corrected.

kamikazechaser:
Updated changes. Documentation reflects a paradox, but most guys should get it 😉

doc/api.md
```diff
@@ -338,6 +339,21 @@ Use this method to send video files, Telegram clients support mp4 videos (other
...
 | [options] | <code>Object</code> | Additional Telegram query options |
 
+<a name="TelegramBot+sendVideoNote"></a>
+
+### telegramBot.sendVideoNote(chatId, videoNote, [options]) ⇒ <code>Promise</code>
+Use this method to send rounded square videos of upto 1 minute long.
+
+**Kind**: instance method of <code>[TelegramBot](#TelegramBot)</code>  
+**See**: https://core.telegram.org/bots/api#sendvideonote  
+**Note**: The length parameter is optional, but **required** untill Telegram patches it.
```
GochoMugo: Is this generated from the source code?

GochoMugo: Please move this into the source code. Otherwise, it will be overwritten once the docs are re-generated.

To be more precise: *The `length` parameter is actually **optional**. However, the API (at time of writing) requires you to **always** provide it  until it is fixed.*

kamikazechaser:
Fixed!

yagop:
Tests were failing because server replied with `video` rather than `video_note`:

```js
{ message_id: 44723,
  from: { ... },
  chat:  { ... },
  date: 1499373920,
  video: 
   { duration: 3,
     width: 560,
     height: 320,
     mime_type: 'video/mp4',
     thumb: 
      { file_id: 'AAQEABOsLIwZAAR7lN4hPBhkCL-XAAIC',
        file_size: 1988,
        width: 90,
        height: 51 },
     file_id: 'BAADBAADMAEAApGS-VKlkzbbFKZ8YQI',
     file_size: 199611 } }
```

Did they change that?

GochoMugo:
I have no clue why it is responding that way. The test seemingly just broke. Is there an issue with the file?

kamikazechaser:
Let me check on it.

kamikazechaser:
It now seems all mp4 files are being sent as videos, other formats are unaffected but not playable. This seems to be an issue on Telegram side.

GochoMugo:
@kamikazechaser Contacted Bot Support?

kamikazechaser:
@GochoMugo They haven't replied yet.

yagop:
Any update on this?

PR #333, opened 2017-05-19
[API] v3 New update types and message sub-types

kamikazechaser:
## API v3

**New Update & Message types**

[[source](https://core.telegram.org/bots/api#update)]
[[source](https://core.telegram.org/bots/api#message)]

Added 2 new update types: `shipping_query`, `pre_checkout_query`

Added 2 new message sub-types: `successful_payment`, `invoice`

- [x] Run tests.

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -20,10 +20,10 @@ const pump = require('pump');
 const deprecate = require('depd')('node-telegram-bot-api');
 
 const _messageTypes = [
-  'text', 'audio', 'document', 'game', 'photo', 'sticker', 'video', 'voice', 'video_note', 'contact',
-  'location', 'new_chat_members', 'left_chat_member', 'new_chat_title',
-  'new_chat_photo', 'delete_chat_photo', 'group_chat_created', 'migrate_to_chat_id', 'migrate_from_chat_id',
-  'supergroup_chat_created', 'channel_chat_created', 'pinned_message', 'invoice', 'successful_payment'
+  'audio', 'channel_chat_created', 'contact', 'delete_chat_photo', 'document', 'game', 'group_chat_created',
+  'invoice', 'left_chat_member', 'location', 'migrate_from_chat_id', 'migrate_to_chat_id', 'new_chat_members',
+  'new_chat_photo', 'new_chat_title', 'photo', 'pinned_message', 'sticker', 'successful_payment', 'supergroup_chat_created',
+  'text', 'video', 'video_note', 'voice'
 ];
 
 // enable cancellation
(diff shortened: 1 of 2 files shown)


GochoMugo:
The code looks great.

However, do we start worrying about polluting the event namespace?

/cc @yagop

src/telegram.js
```diff
@@ -20,9 +20,10 @@ const pump = require('pump');
 const deprecate = require('depd')('node-telegram-bot-api');
 
 const _messageTypes = [
-  'text', 'audio', 'document', 'photo', 'sticker', 'video', 'voice', 'contact',
-  'location', 'new_chat_participant', 'left_chat_participant', 'new_chat_title',
-  'new_chat_photo', 'delete_chat_photo', 'group_chat_created'
+  'text', 'audio', 'document', 'game', 'photo', 'sticker', 'video', 'voice', 'video_note', 'contact',
```
GochoMugo: Let's just arrange this alphabetically, for less cognitive load for a user looking through source code.

GochoMugo:
Another issue we need to address is the fact that some Message properties are (sic. will be) removed thus their events will **not** be fired, which *might* be viewed as breaking the API. Also, doesn't the renaming of the event above break our API?

/cc @yagop

src/telegram.js
kamikazechaser: Ok

kamikazechaser:
Which ones?

GochoMugo:
> Replaced the field `new_chat_member` in Message with `new_chat_members` (the old field will still be available for a while for compatibility purposes).
> 
> &mdash; https://core.telegram.org/bots/api-changelog#may-18-2017
>
> Renamed the fields `new_chat_participant` and `left_chat_participant` of the Message object to `new_chat_member` and `left_chat_member`.
>
> &mdash; https://core.telegram.org/bots/api-changelog#april-9-2016

This might be a non-issue really. But how should we ensure we do **not** do something unexpected by the user. Should we provide "polyfills" on top, to hide the API updates (until we decide to bump major!), that allow the library to work as expected; emit the expected events?

kamikazechaser:
@GochoMugo I had made the necessary changes. The types are up to date.

GochoMugo:
Let's split this PR into two:

PR 1:

* new update types: `shipping_query`, `pre_checkout_query`
* new message sub-types: `successful_payment`, `invoice`

PR 2:

* missing message sub-types: `game`, `pinned_message`, `migrate_from_chat_id`, `migrate_to_chat_id`, `channel_chat_created`, `supergroup_chat_created`
* rename 1 message sub-type: `new_chat_member`

That would allow us to address the issues mentioned above properly.

kamikazechaser:
@gochomugo Since I have alphabetically sorted it, it would mean we would have to discuss and merge in PR #340 first before merging in this one.

GochoMugo:
You could undo the sorting, and simply append the new events. We shall do sorting (etc.) in #340.

yagop:
Seems there are many events but its ok IMO ...

kamikazechaser:
@GochoMugo Reverted to the original arrangement.

PR #335, opened 2017-05-23
[API] v3 Payments

kamikazechaser:
## API v3 

**sendInvoice**

[[source](https://core.telegram.org/bots/api#sendinvoice)]

- [x] Add tests.
- [x] Generate documentation.
- [x] Run tests.

**answerShippingQuery**

[[source](https://core.telegram.org/bots/api#answershippingquery)]

- [x] Generate documentation.

**answerPreCheckoutQuery**

[[source](https://core.telegram.org/bots/api#answerprecheckoutquery)]

- [x] Generate documentation.

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -1255,6 +1255,22 @@ class TelegramBot extends EventEmitter {
     form.prices = JSON.stringify(prices);
     return this._request('sendInvoice', { form });
   }
+
+  /**
+   * Answer shipping query..
+   * Use this method to reply to shipping queries.
+   *
+   * @param  {String} shippingQueryId  Unique identifier for the query to be answered
+   * @param  {Boolean} ok Specify if delivery of the product is possible
+   * @param  {Object} [options] Additional Telegram query options
+   * @return {Promise}
+   * @see https://core.telegram.org/bots/api#answershippingquery
+   */
+  answerShippingQuery(shippingQueryId, ok, form = {}) {
+    form.shipping_query_id = shippingQueryId;
+    form.ok = ok;
+    return this._request('answerShippingQuery', { form });
+  }
 }
 
 module.exports = TelegramBot;
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -1271,6 +1271,22 @@ class TelegramBot extends EventEmitter {
     form.ok = ok;
     return this._request('answerShippingQuery', { form });
   }
+
+  /**
+   * Answer pre-checkout query.
+   * Use this method to confirm shipping of a product.
+   *
+   * @param  {String} preCheckoutQueryId  Unique identifier for the query to be answered
+   * @param  {Boolean} ok Specify if every order details are ok
+   * @param  {Object} [options] Additional Telegram query options
+   * @return {Promise}
+   * @see https://core.telegram.org/bots/api#answerprecheckoutquery
+   */
+  answerPreCheckoutQuery(preCheckoutQueryId, ok, form = {}) {
+    form.pre_checkout_query_id = preCheckoutQueryId;
+    form.ok = ok;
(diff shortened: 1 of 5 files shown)


src/telegram.js
```diff
@@ -1167,6 +1167,34 @@ class TelegramBot extends EventEmitter {
...
+   */
+  sendInvoice(chatId, title, description, payload, providerToken, startParameter, currency, prices, form = {}) {
+    form.chat_id = chatId;
+    form.title = title;
+    form.description = description;
+    form.payload = payload;
+    form.provider_token = providerToken;
+    form.start_parameter = startParameter;
+    form.currency = currency;
+    form.prices = JSON.stringify(prices);
```
kamikazechaser: @GochoMugo I prefer we stringify the array here, rather than make people do it in their bot. Similar to the way we do it with anserInlineQuery. What do you think?

GochoMugo: That's the our job in this library :laughing: Yes, we should certainly do it for the users! :+1:

GochoMugo:
> Replaced the field `new_chat_member` in Message with `new_chat_members` (the old field will still be available for a while for compatibility purposes).
> 
> &mdash; https://core.telegram.org/bots/api-changelog#may-18-2017
>
> Renamed the fields `new_chat_participant` and `left_chat_participant` of the Message object to `new_chat_member` and `left_chat_member`.
>
> &mdash; https://core.telegram.org/bots/api-changelog#april-9-2016

This might be a non-issue really. But how should we ensure we do **not** do something unexpected by the user. Should we provide "polyfills" on top, to hide the API updates (until we decide to bump major!), that allow the library to work as expected; emit the expected events?

src/telegram.js
GochoMugo: Is this necessary? Does it fail if we do **not** stringify?

test/telegram.js
```diff
@@ -20,6 +20,7 @@ if (!TOKEN) {
 const USERID = process.env.TEST_USER_ID || 777000;
 const GROUPID = process.env.TEST_GROUP_ID || -1001075450562;
 const GAME_SHORT_NAME = process.env.TEST_GAME_SHORT_NAME || 'medusalab_test';
+const PROVIDER_TOKEN = process.env.TEST_PROVIDER_TOKEN;
```
GochoMugo: Until we formulate a way to run a subset of the tests, depending on the available configurations, we should check if this token is provided (as we do with the `TEST_TELEGRAM_TOKEN` above), i.e.

```js
if (!PROVIDER_TOKEN) {
   throw new Error('Payment provider token not provided');
}
```

src/telegram.js
kamikazechaser: Yup, it fails.

kamikazechaser: Unless one forms a valid JSON with quotes on his side. The stringify makes things easier at no extra cost.

GochoMugo: But the documentation says nothing about this? :astonished:

kamikazechaser: Yeah it doesn't tbh. But just like we stringify on https://github.com/yagop/node-telegram-bot-api/blob/master/src/telegram.js#L554, it would make it less tedious to form an array of objects.

GochoMugo: But the docs (https://core.telegram.org/bots/api#answerinlinequery) instructs user to serialize the parameter into JSON...

kamikazechaser: I guess the same applies for payments, but they havent stated. I tested it and it seems to work when serialized.

kamikazechaser:
@GochoMugo This seems ok.

GochoMugo:
@yagop You'll need to add the Payments provider token, i.e. `TEST_PROVIDER_TOKEN` for this to fix our tests.

PR #340, opened 2017-05-24
[Update] Message sub-types

kamikazechaser:
[[source](https://core.telegram.org/bots/api#message)]

Added 6 missing message sub-types: `game`, `pinned_message`, `migrate_from_chat_id`, `migrate_to_chat_id`, `channel_chat_created`, `supergroup_chat_created`

Rename 1 message sub-type: `new_chat_members`

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -14,9 +14,12 @@
   1. Depending on the properties of the [Message][message], one of these
      events will **ALSO** be emitted: `text`, `audio`, `document`, `photo`,
      `sticker`, `video`, `voice`, `contact`, `location`,
-     `new_chat_participant`, `left_chat_participant`, `new_chat_title`,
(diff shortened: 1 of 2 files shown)


GochoMugo:
What will happen if a user listens on the old events (`new_chat_participant`, `left_chat_participant`)? They'll **not** receive the events, as they expect to, despite us guaranteeing that we have **not** broken our (library) API. I suggest we throw an error if the user listens on these old events.

kamikazechaser:
@GochoMugo I think we should start supporting the new message sub-types, since they were released in v2.3.3. People have to adapt to changing situations!

GochoMugo:
I have added some logic to log a deprecation notice should the user listen on events `new_chat_participant` and `left_chat_participant`.

kamikazechaser:
@GochoMugo Perfect!

PR #413, opened 2017-08-24
Expose message.type to 'message'

### LINKED ISSUE #409: Expose the message type to on('message')
> Feature Request

I have:

* searched for such a feature request (https://github.com/yagop/node-telegram-bot-api/labels/enhancement) and found none

### Introduction

Especially useful for logging/tracking purposes, expose the message type to the `message` event, avoiding code duplication just to find out the message type

### Example

```js
bot.on('message', function onMessage(msg) {
  botan.track(msg, (msg.text)?msg.text.split(' ')[0]:msg.type);
}
```
(example using botanio, but would be applicable to pretty much any tracking library/system, the msg.text part is added as a more complete example, to track commands)

jlsjonas:
- [ ] All tests pass
- [ ] I have run `npm run gen-doc`

(haven't run any as it was a trivial fix directly from the github interface; I'd be highly surprised if things don't work out)

### Description

expose `message.type` to the 'message' event, the checks are being done already either way

### References

Closes #409

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -491,11 +491,13 @@ class TelegramBot extends EventEmitter {
 
     if (message) {
       debug('Process Update message %j', message);
-      this.emit('message', message);
       const processMessageType = messageType => {
         if (message[messageType]) {
           debug('Emitting %s: %j', messageType, message);
           this.emit(messageType, message);
+          this.emit('message', { ...message, type: messageType });
+        } else {
+          this.emit('message', message);
         }
       };
       TelegramBot.messageTypes.forEach(processMessageType);


src/telegram.js
```diff
@@ -491,11 +491,13 @@ class TelegramBot extends EventEmitter {
...
     if (message) {
       debug('Process Update message %j', message);
-      this.emit('message', message);
       const processMessageType = messageType => {
         if (message[messageType]) {
           debug('Emitting %s: %j', messageType, message);
           this.emit(messageType, message);
+          this.emit('message', { ...message, type: messageType });
+        } else {
+          this.emit('message', message);
```
GochoMugo: I see a bug! Considering that the function `processMessageType` in invoked several times (below: `TelegramBot.messageTypes.forEach(processMessageType);`), this branch will cause the `message` event to be emitted in any case where the message is of the type being inspected!

Considering that a message is of **exactly** one type e.g. `audio`, `text`, etc, we could first find the type. For example,

```js
const messageType = TelegramBot.messageTypes.find((type) => message[type]);
message.type = messageType;
```
Then we can now emit the events:

```js
this.emit("message", message);
this.emit(messageType, message);
```
Using `.find()` will actually improve our performance.

GochoMugo:
Superseded by commit 49df0c6e0233124a9a24bae23e4ca93773a4f69a.

Issue #448, opened 2017-10-25
answerShippingQuery calls Error


dobeerman:
I'm trying to make an `answerShippingQuery` with code example below, but something went wrong. 
I've the following error:

> Unhandled rejection Error: ETELEGRAM: 400 Bad Request: shipping options are not specified

```
bot.on("shipping_query", shipping => {
  const ok = true;
  const opts = {
    shipping_options: [
      {
        id: "shipping",
        title: "Shipping by seller",
        prices: [
          { label: "Shipping by seller 1", amount: 500 },
          { label: "Shipping by seller 2", amount: 500 }
        ]
      },
      {
        id: "takeaway",
        title: "Take away",
        prices: [{ label: "-Take away", amount: 0 }]
      }
    ],
    error_message: ""
  };

  bot.answerShippingQuery(shipping.id, ok, opts);
});
```

drGOD:
I have the same problem

GochoMugo:
Fixed in **v0.30.0**.

PR #468, opened 2017-11-25
Fix issue with answerShippingQuery

drGOD:
- [x] All tests pass

### Description

Fix issue #448 with `answerShippingQuery`. The problem was, that `shipping_options` doesn't converting to JSON from array

### References

* Issue #448


GochoMugo:
Thanks @drGOD for the fix! :+1:

PR #523, opened 2018-02-14
sendMessage: pass valid parameters to API

gurugray:
perhaps after 13.02.2018 API update there are only errors like "**400: message are empty**" when we are trying to send a message via `sendMessage` method

this fix the issue.

- [ ] All tests pass

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -672,10 +672,10 @@ class TelegramBot extends EventEmitter {
    * @return {Promise}
    * @see https://core.telegram.org/bots/api#sendmessage
    */
-  sendMessage(chatId, text, form = {}) {
-    form.chat_id = chatId;
-    form.text = text;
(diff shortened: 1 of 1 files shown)


PashaGeronimo:
Me too!

gurugray:
@yagop please check this fix — current work of `node-telegram-bot-api` is broken :(

GochoMugo:
Any steps in reproducing this bug?

gurugray:
I can't reproduce it anymore.
But in time of reporting — there were no successful message sending at all :\

GochoMugo:
@PashaGeronimo How about you? Able to reproduce this?

kamikazechaser:
It has been noted several times in the past that Telegram servers may respond/behave weirdly, especially when they are rolling out new features to the API. I will close this pr since it cannot be reproduced anymore.

PR #583, opened 2018-05-13
set request options in download file

khodadadi:
fix bug


src/telegram.js
```diff
@@ -1379,7 +1379,12 @@ class TelegramBot extends EventEmitter {
         fileStream.emit('info', {
           uri: fileURI,
         });
-        pump(streamedRequest({ uri: fileURI }), fileStream);
+        let options = {};
```
GochoMugo: Please use `const` in place of `let`.

kamikazechaser:
Closed in #771

Issue #585, opened 2018-05-13
Bot replies 401 Unauthorized to webhooks if the bot token is not part of the webhook URL


LeartS:
> Bug Report

I have read:

* [Usage information](https://github.com/yagop/node-telegram-bot-api/tree/master/doc/usage.md)
* [Help information](https://github.com/yagop/node-telegram-bot-api/tree/master/doc/help.md)

I am using the latest version of the library.

### Summary

The bot replies 401 to webhooks from telegram if the bot token is not parth of the webhook url.
[This seems to be an explicit choice](https://github.com/yagop/node-telegram-bot-api/blob/8c965d3a14a8941e35f4c60f00c4cd2e55d17038/src/telegramWebHook.js#L134)

### Expected Behavior
The bot replies to webhooks normally

### Actual Behavior

The bot answers with 401 Unauthorized if the bot token is not present in the path passed to `setWebhook`. This is 

The telegram docs *recommends*, but not mandates, using a secret string in the webhook url, and suggests the bot token just as an example. But any randomly-generated string is ok for the purpose; i.e. have an update url that only the developer and telegram know.

IMO the limitation should be removed or, at the very least, should be documented. I had to check the source code to see why the bot was answering 401 to the webhook notifications.

### Steps to reproduce the Behavior
Use `setWebhook` to configure webhooks at a path that doesn't contain the bot token. e.g. 

    bot.setWebhook('https://mybotdomain.com/webhook/random-secret-string-fjfjkhbf84bg8754')

GochoMugo:
@LeartS Thanks for taking your time to report this issue. Yes, I agree. We should **not** force the token to be used in the webhook URL. Are you willing to send a PR our way adding this enhancement? We could allow an additional, optional parameter specifying such a random token in the webhook URL. Documentation could be improved too.

PR #614, opened 2018-07-09
Implement webHook endpoint

LadnovSasha:
- [x] All tests pass

### Description
`TOKEN` webHook endpoint is now optional.

You can specify `secret path` in options
```
new TelegramBot(TOKEN,{
  webHook: {
    endpoint: 'your_secret_path' //Defaults to index path
  }
});
```
### References
https://github.com/yagop/node-telegram-bot-api/issues/585

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -59,6 +59,7 @@ $ openssl genrsa -out key.pem 2048
 # Our public certificate will be crt.pem
 $ openssl req -new -sha256 -key key.pem -out crt.pem
 ```
+**Note:** Domain, specified in CN section during creation should be equal to your real domain address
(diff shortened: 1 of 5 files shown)


kamikazechaser:
Leaving this comment for future reference:

This approach is not as flexible as something like [this](https://github.com/yagop/node-telegram-bot-api/blob/master/examples/webhook/express.js) which allows for finer route level flexibility and independence of setting up certificates.

So you would probably have an Caddy/Nginx server with a cert already configured, you then proxy pass it to your express web server and process updates as usual.

If this is a much needed feature, it can be re-opened, but I doubt that.

PR #668, opened 2018-10-24
Send thumb with documents

mohammadrezaprw:
- [ ] All tests pass


kamikazechaser:
This is no longer needed. Telegram will automagically generate a thumb for `sendDocument`.

PR #678, opened 2018-11-04
Added node-telegram-operation-manager to community field

alexandercerutti:
I've created this utility library for bots, that uses this library to show usage example. 😊


kamikazechaser:
Great work!

PR #746, opened 2019-09-09
Added various type definitions

adityathebe:
- TelegramUser
- TelegramLocation
- TelegramInlineQuery
- TelegramChosenInlineResult
- TelegramCallbackQuery
- TelegramShippingAddress
- TelegramShippingQuery
- TelegramPollOption
- TelegramPoll
- TelegramOrderInfo
- TelegramPreCheckoutQuery
- TelegramUpdate
- TelegramMessage [Incomplete]

- [ ] All tests pass


kamikazechaser:
I don't think JSDoc2Markdown will pick up `typedef`` definitions by default from the src code unless we explicitly include it. Moreover, these typedefs are aalready in the official TG docs. So what exact purpose would it serve? Maybe a `*.d.ts` (TypeScript def) would be useful.

adityathebe:
Type definitions already exist

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node-telegram-bot-api/index.d.ts

PR #783, opened 2020-03-21
fix regexp compare

ledamint:
- [x] All tests pass

### Description

Fix bug with compare two regexp
```javascript
/123/ === /123/ // false
String(/123/) === String(/123/) // true
```


ledamint:
@kamikazechaser Please take a look. `removeTextListener` doesn't work at all.

kamikazechaser:
You are right, the comparison will always  fail because /a/ is an object type!

PR #824, opened 2020-08-13
How to set webhooks using express local server and NGROK

leobloise:
It's a guide to help new people how to set webhooks and test their bots using local server.


kamikazechaser:
Thanks for the tutorial!

PR #834, opened 2020-10-27
Add methods to clear reply and text listeners

daniel0611:
- [x] All tests pass

### Description

Sometimes it can be handy to be able to remote all reply and text listeners. Also removes the burden of needing to track what listeners are registered so you can unregister them with `removeTextListeners`/`removeReplyListeners`. Instead you can just use `clearTextListeners`/`clearReplyListeners`.


kamikazechaser:
Thanks. useful methods.

PR #843, opened 2020-12-08
fix: send animation

satouriko:
- [ ] All tests pass

### Description

Send Animation requires an `animation` field instead of `document`.

### References

https://core.telegram.org/bots/api#sendanimation


kamikazechaser:
Thanks for spotting this.

PR #846, opened 2020-12-28
setStickerSetThumb function added

elihaidv:
https://core.telegram.org/bots/api#setstickersetthumb

Now you can add thumb image to the sticker set

- [ ] All tests pass


kamikazechaser:
Could you add some tests for this?

PR #853, opened 2021-01-27
Adding thumb support for for supported methods

0tabek16:
- [x] All tests pass

### Description

Ability send `thumb`s for supported methods.

### References

https://core.telegram.org/bots/api#sendaudio

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -240,6 +240,24 @@ class TelegramBot extends EventEmitter {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kamikazechaser:
This is probably not needed. Telegram generates the thumbnails server side if possible and **ignores** the thumb param if it can generate it. Maybe in some edge cases you will need to give your own thumb. Can you provide a sample file hosted somewhere? Almost all the files I am testing have a thumbnail.

PR #877, opened 2021-05-13
Added support for listening on UNIX sockets for webhook requests

exactly-one-kas:
- [ ] All tests pass
  - I don't have a provider token to run the tests

### Description

This PR allows bot admins to use webhooks with UNIX sockets to bind for reverse proxies

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -132,7 +132,7 @@ Emits `message` when a message arrives.
 | [options.webHook] | <code>Boolean</code> \| <code>Object</code> | <code>false</code> | Set true to enable WebHook or set options |
 | [options.webHook.host] | <code>String</code> | <code>&quot;0.0.0.0&quot;</code> | Host to bind to |
 | [options.webHook.port] | <code>Number</code> | <code>8443</code> | Port to bind to |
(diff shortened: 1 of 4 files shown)


kamikazechaser:
Definitely interesting to listen on UNIX sockets. However what happens if a user sets both `port` and `path`? From what I can see both Promises get pushed. Also rename `path` to `socketPath` for semantic reasons.

exactly-one-kas:
> However what happens if a user sets both `port` and `path`?

It should listen both on TCP as well as UNIX sockets. Is that acceptable or should I refactor it like you wrote?
Personally, this isn't a use case for me.

> Also is there any real benefit of using UNIX sockets? […] Does it really save on resources?

For me it's about neatness, since I don't have to check for a free port or remember which app runs on which port.

kamikazechaser:
Yes, I believe its better to refactor it since we this library has a lot of Windows users also some users may open issues asking why a server is listening on both a UNIX socket and a TCP/IP port. Also I guess you won't need to promisify all since only 1 branch of the conditional will be taken with the refactor I suggested.

PR #897, opened 2021-08-15
renew

mohammadrezaprw:
- [ ] All tests pass


mohammadrezaprw:
upstream

kamikazechaser (commented):
1. We don't need type definitions here. They are being maintained in another repo.
2. Removing gen-doc will break docs

kamikazechaser:
Also is this regarding a TG side update? I may have missed it. Let me know the details.

PR #907, opened 2021-09-29
restrictChatMember add untilDate

CreatorOfEverything159:


danielperez9430:
Hi, I wanted to tell you that this pull request has a couple of problems:

- Don't add `.idea /*` files in commit (That are files from your visual studio)
- That method function works perfectly, so no need any change
- `until_date` is a **optional** parameter (so you can't add it as a option)
- The third parameter "form = {}" is optional and you need to pass a object with options

Example of use *restrictChatMember*:

```js
let time = 30; // Time in seconds

// Time in ms - unixtime
let date = Math.round(new Date(Date.now()+time)/1000);

let options = {
  can_send_messages:false,
  can_send_other_messages: false,
  can_add_web_page_previews: false,
  can_pin_messages: false,
  can_change_info: false,
  can_invite_users: false,
  can_send_polls: false,
  can_send_media_messages: false,
  can_can_add_web_page_previews: false
}

options.until_date = date

bot.restrictChatMember(msg.chat.id, msg.from.id, options);
```

Issue #223, opened 2016-11-10
Message is too long "error_code":400


EllysahatNMI:
Hi! I was using the simple "bot.sendMessage" api to send messages but I received this error. 

Do you know the limit to the length of the message? Or maybe tell me how I can change the limit.

GochoMugo:
The Maximum number of characters allowed is 4096! You will have to split your messages. (Check out [tgfancy](https://github.com/GochoMugo/tgfancy) for a possible solution)

Issue #19, opened 2015-08-13
Unhandled rejection Error: 414 Request-URI Too Large


CTassisF:
Hello.

I think when the message sent using sendMessage is too long it triggers the following error:

```
<html>
<head><title>414 Request-URI Too Large</title></head>
<body bgcolor="white">
<center><h1>414 Request-URI Too Large</h1></center>
<hr><center>nginx/1.6.2</center>
</body>
</html>

    at /home/services/node/telegram/node_modules/node-telegram-bot-api/src/telegram.js:85:15
    at tryCatcher (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:577:18)
    at Promise._settlePromises (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/promise.js:693:14)
    at Async._drainQueue (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:133:10)
    at Async.drainQueues (/home/services/node/telegram/node_modules/node-telegram-bot-api/node_modules/bluebird/js/main/async.js:15:14)
    at process._tickCallback (node.js:419:13)
```

The bot stops working after that.

Is there any way to fix this issue (and/or report it directly to Telegram team to enable bots to send longer messages)?

Thanks in advance.

yagop:
As you said, it seems the message is too large to be sended. How long is it?
Maybe you should contact Telegram [bot support](telegram.me/BotSupport).

CTassisF:
I will ask them what is the current size limit of a single message.

Anyway, can you please update node-telegram-bot-api to deal with this kind of error without entering some sort of "vegetative state" (as it currently does)?

Thanks in advance.

SerjoPepper:
This bug is not related with telegram message size. It is related with GET request uri size restrictions http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.15
[...]

PR #992, opened 2022-08-08
Fix message too long

Daniel-97:
- [x] All tests pass

### Description

This PR try to solve the issue with the error "Message too long" when sending message larger than 4096 characters.
I think this is somthing that the library should handle and don't let the user deal with it.

### References

* Issue #223 https://github.com/yagop/node-telegram-bot-api/issues/223
* Issue #19  https://github.com/yagop/node-telegram-bot-api/issues/19


larchanka (approved):

Daniel-97:
I have generated the new doc because i have added some line to the documentation comment of the sendMessage() function. Let me know if it is ok. 
Love your work!

kamikazechaser:
Tbh this is outside the scope of NTBA. NTBA is meant to be a low level library. Such a feature should be handled by user level code  that guarantees order, retry and memory safety. Also `async/await` breaks the current code style of this library.

You can tag me on the Telegram group if you want to discuss more.

PR #1006, opened 2022-09-14
Changed example commands regex

seadog007:
- [ ] All tests pass (Didn't run it since I am not change the lib source)

### Description
Changed the examples and README.md since lots of people wrote codes with copying the example, which might cause the bot to match URL or other things. 

Generally, the only thing that should consider as a command is a string start with special character (in this case "/")
Therefore, the following should be considering as a command
```
/echo 123
/help
```

but not these
```
I found a file under /bin, which is /bin/echo What is is file for in Linux? 
http://help.google.com/
```

Using the original example will resulting as
截圖 2022-09-15 上午1 19 42

### References
Nope

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ const token = 'YOUR_TELEGRAM_BOT_TOKEN';
 const bot = new TelegramBot(token, {polling: true});
 
 // Matches "/echo [whatever]"
-bot.onText(/\/echo (.+)/, (msg, match) => {
+bot.onText(/^\/echo (.+)/, (msg, match) => {
   // 'msg' is the received Message from Telegram
   // 'match' is the result of executing the regexp above on the text content
   // of the message
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ const token = 'YOUR_TELEGRAM_BOT_TOKEN';
 const bot = new TelegramBot(token, {polling: true});
 
 // Matches "/echo [whatever]"
-bot.onText(/^\/echo (.+)/, (msg, match) => {
+bot.onText(/^\/echo(?:@.*?)? (.+)/, (msg, match) => {
   // 'msg' is the received Message from Telegram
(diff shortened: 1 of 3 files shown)


kamikazechaser:
You probably also want to add the regex terminator to some of these examples i.e. $

seadog007:
@kamikazechaser I did consider that, but since there might be something like `/command@botname`, and regex will be `^\/command(@.*|$)`.
I don't think complexing the example is a good idea then.
Therefore, I was not adding that to the PR, or should I?

kamikazechaser:
> /command@botname

This is usually in groups. Yeah I think you should add it for uniformity.

danielperez9430:
This regex is still wrong, becuase @.* is any character include espace and other text that not match.

`/test@mytest_bot` and '/test@mytest_bot hello test' is match.

This weekend I will review it

seadog007:
@danielperez9430 Should be `^\/echo(?:@.*?)?( .*|$)`
That's my mistake

seadog007:
@danielperez9430 
Well, I will simply ignore the argument for the commands that doesn't require args.
but for something like `/echo`, I did use `^\/echo(?:@.*?)? (.+)`, or you prefer `^\/echo(?:@.*?)?( .*|$)` ?

kamikazechaser:
> Yeah I think you should add it for uniformity.

Btw I meant the regex terminator here, not adding `@botname` to the regex. Otherwise the PR is okay after that change.

yagop:
Thanks for catching this back in 2022, @seadog007 — the point is still valid: the unanchored example regexes match `/echo` (and friends) anywhere in a message, including inside URLs or mid-sentence. Your original patch predated the TypeScript rewrite and no longer applies to the current files, so I've reimplemented the fix fresh against today's `examples/` + `README.md` in #1311 (anchoring each example command with a leading `^`, with a short note explaining why).

Closing this in favor of #1311 — credited to you. Appreciate the report!

Issue #1015, opened 2022-10-21
Sending long messages


Polm90:
Hi,
if you try to send a long message (over 4000 characters) you receive a 400 error from the telegram server for a too long message.
This is something which is not (well) documented (I think it should be written better in the docs).

Can someone suggest a smart way to handle this (ie splitting in the right way the message, without cutting html tags, and sending each part subsequently?

However, it would be nice if the libraries could handle directly this, allowing the user to send too long messages in the right way :)

Thank you!

PR #1052, opened 2023-02-13
add sendLongMessage function to telegram.js

shahabhm:
- [ ] All tests pass

### Description
I have added a method to handle sending long messages. this method splits the text into smaller parts and sends them. I also can add a test if you agree with this idea. 

### References
this resolves the following issue:
[Sending long messages](https://github.com/yagop/node-telegram-bot-api/issues/1015)

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -915,6 +915,38 @@ class TelegramBot extends EventEmitter {
   }
 
   /**
+   * Sends long text message.
+   * currently, telegram bots can send only 4000 characters per message. to send a message longer than this length,
+   * you need to split it into multiple messages and then send them separately. this function does that.
+   * @param  {Number|String} chatId Unique identifier for the target chat or username of the target channel (in the format `@channelusername`)
+   * @param  {String} text Text of the message to be sent
+   * @param  {Object} [options] Additional Telegram query options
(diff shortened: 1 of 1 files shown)


kamikazechaser:
Well tbh, such a feature alongside #997 should not live in this library. It could either be a util or live outside the library.

One reason is that a quick fire of very long messages e,g, 50k characters+ can easily lead to 429's which can break the correct sequence.

shahabhm:
> 

that's right. do you think adding a limit to the text size is enough? like 8k or 12k characters?

ArtToledo:
any solution about this? I have the same problem

ArtToledo:
image
image

I found an error in the implementation when using Math.Max, considering text.length as the second value, so the first time you run chunck, the idea is to cut the message into 4000 characters, with math.max in messages above this an X value, in my case 7741 characters, it gets the wrong chunck_end considering the total size of the message instead of considering where it should break the string

GochoMugo:
As noted earlier by @kamikazechaser, this PR is out of scope for this library. Please consider sharing this change as a separate library/package. For example,

```ts
const sendLongMessage = require("telegram-send-long-message");

await sendLongMessage(bot, myVeryLongMessage);
```

Nonetheless, thanks @shahabhm for this PR.

PR #1108, opened 2023-07-27
fixed unpinChatMessageBug

serjou1:
- [x] All tests pass

### Description

unpinChatMessage method had a bug, it didn't use messageId properly so I got error
`Failed to unpin message: TypeError: Cannot create property 'chat_id' on number <message id>`


kamikazechaser:
It is an optional param. Use it like:

```js
bot.unpinChatMessage(GROUPID, {message_id: xxx})
```

PR #1121, opened 2023-09-12
Fix caption_entities stringify in sendAudio

Sp3rick:
caption_entities option in sendAudio was not automatically stringified

- [ ] All tests pass

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -1008,6 +1008,10 @@ class TelegramBot extends EventEmitter {
     } catch (ex) {
       return Promise.reject(ex);
     }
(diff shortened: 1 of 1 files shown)


Sp3rick:
I close because i see that's miss in every function with option caption_entities, should be done something like _fixReplyMarkup, or copy paste this:
`if( opts.qs.hasOwnProperty( "caption_entities" ) && Array.isArray(opts.qs.caption_entities) )
    {
      opts.qs.caption_entities = stringify(opts.qs.caption_entities);
    }`

for every function with caption_entities option (yes, i added array check)

Sp3rick:
moved to #1122

PR #1122, opened 2023-09-12
added _fixCaptionEntities

Sp3rick:
Sent caption_entities messages are not stringified by default, this function will do that

I implemented it ONLY in: sendVoice sendAnimation sendVideo sendDocument sendAudio and sendPhoto

May be missing copyMessage and all edit functions

- [ ] All tests pass

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -252,6 +252,11 @@ class TelegramBot extends EventEmitter {
       }
     }
   }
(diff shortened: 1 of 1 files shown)


kamikazechaser:
caption_entities are an optional parameter. It is up to the user to handle ir.

danielperez9430:
Yep, optional params need to handle manual by the user in their code

Issue #1076, opened 2023-04-20
request-promise is deprecated


microcai:
tgbot is using request-promise as https client. but request-promise is deprecated by upstream developers.
see https://github.com/request/request/issues/3143

hodlerhacks:
In addition, it uses 'request', which has been deprecated. And 'request' has a dependency on 'har-validator', which is no longer supported, and on an a deprecated version of 'uuid'.

It would be great if this could be resolved by moving to an alternative, e.g. see some examples here: https://github.com/request/request/issues/3143

Other than that, I'm super happy with this library. Thanks for the great work!

muety:
Since version 18 (I believe) Node has the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) included, so no more need for a third-party HTTP client.

hodlerhacks:
Indeed. The only disadvantage is that it requires every application using tgbot to be on Node 18 or above (which may cause problems for some people, including me ;-)).

jinsankim:
how about [node-fetch](https://www.npmjs.com/package/node-fetch)? It's a polyfill of `fetch` and it requires only node 12.20.0 at least.

melroy89:
Not only is `request` / `request-promise` deprecated, now you also got a nasty server side request forgery attack security issue: https://github.com/advisories/GHSA-p8p7-x288-28g6 ... so the urgency is just become more important to move away from this package!

melroy89:
And this package now also causes issues with bun!

Hambat:
any updates?

danielperez9430:
fix: on next release

jinsankim:
@cypress/request is not working with node v20.x. :(

danielperez9430:
> @cypress/request is not working with node v20.x. :(

I test in node v20.9.0 and i dont have any problem 🤔

At less you have one bot.onText('... 
with a text that is not a regex

jinsankim:
> I test in node v20.9.0 and i dont have any problem

I found also it works in node v20 with mac. But it doesn't work in node v20 with `ubuntu`.

saeedhei:
Does this bot have the ability to update with axios?
@danielperez9430 @kamikazechaser @yagop

danielperez9430:
> Does this bot have the ability to update with axios?
> 
> @danielperez9430 @kamikazechaser @yagop 

[...]

PR #1125, opened 2023-09-23
chore: Replace request to node-fetch, #1076

Dima-Dim:
- [x] All tests pass

About tests:
I don't have TEST_STICKER_SET_NAME and TEST_PROVIDER_TOKEN to run the tests related to them. The rest of the test pass normally.

### Description
Getting rid of deprecated dependencies request.

### References
https://www.npmjs.com/package/request

Issue: #1076.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
     "eventemitter3": "^3.0.0",
     "file-type": "^3.9.0",
     "mime": "^1.6.0",
+    "node-fetch": "2.7.0",
     "pump": "^2.0.0",
-    "request": "^2.83.0",
     "request-promise": "^4.2.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


kamikazechaser:
Looks like a good change if all tests pass, though Node.js has a built in fetch API in newer version (18+).

Dima-Dim:
> Looks like a good change if all tests pass, though Node.js has a built in fetch API in newer version (18+).

Yes, after some time it will be possible to try to completely switch to a native solution, but i think no before end of support version older than 18.

0x114514BB:
Your change literally removed the `request` dependency. The core `_request` function of NTBA depends on `request-promise` (also deprecated), which has `request` [as a peer-dependency](https://www.npmjs.com/package/request-promise-core). So you should replace it with `node-fetch` or other solutions.

danielperez9430:
For the moment we have already a remplace for the request library merge and without need to drop support for the current old versions.

In a future completed refactor, we will be drop the support for older versions of node. But for the moment we still preserve the compatibility.

PR #1142, opened 2023-10-19
revokeChatInviteLink api doc fix

matvejs16:
- [x] All tests pass

### Description

Fixed api error for revokeChatInviteLink

### References

No references. This api doc does not match [the actual code](https://github.com/yagop/node-telegram-bot-api/blob/595cdbd0c656be258bebda2aab5d343de1e938d4/src/telegram.js#L1616), which takes inviteLink: string as a second argument

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1113,7 +1113,7 @@ The bot **must be an administrator in the chat** for this to work and must have 
 
 <a name="TelegramBot+revokeChatInviteLink"></a>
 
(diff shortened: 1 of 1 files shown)


danielperez9430:
The API doc file is a file that you can't modify manual. All the time is regenerate (Using the command `npm run doc`). And this need to be change in the file "src/telegram.js" method signature and after that run the command.

I push and merge the correct fix in this commit: https://github.com/yagop/node-telegram-bot-api/commit/f17e801b8a0a194ced929abdcb23e245b953f687

Issue #1111, opened 2023-08-14
tough-cookie@2.5.0 vulnerability issues


ArvinCS:
> Feature Request

I have:

* searched for such a feature request (https://github.com/yagop/node-telegram-bot-api/labels/enhancement) and found none

I think the library needs to upgrade its dependency that dependent to tough-cookie@2.5.0. Since tough-cookie@2.5.0 has vulnerability issue (https://avd.aquasec.com/nvd/cve-2023-26136), it needs to be updated to tough-cookie@4.0.0.

  node-telegram-bot-api@0.61.0
  ├─┬ request-promise@4.2.6
  │   └── tough-cookie@2.5.0
  └─┬ request@2.88.2
       └── tough-cookie@2.5.0

### Introduction

It will solve this vulnerability issue: https://avd.aquasec.com/nvd/cve-2023-26136

### Example

Update the library request and request-promise.

melroy89:
Related: https://github.com/yagop/node-telegram-bot-api/issues/1076

danielperez9430:
fix: on next release

Issue #1038, opened 2022-12-27
[Feature] Replace package request


DreamOfIce:
Request has been deprecated for a long time and it is recommended to switch to other similar libraries: e.g. `axios` or `node-fetch`
https://github.com/request/request/issues/3142

lrobot:
+1, and also can think about,  https://github.com/sindresorhus/got  (a request alternative).

ultrabos1429:
💓

BarakAlmog:
🙌

infinitewhileloop:
I switched to [Telegraf](https://github.com/telegraf/telegraf) because of this issue.

danielperez9430:
fix: on next release

PR #1145, opened 2023-10-24
deps: Change request to @cypress/request

danielperez9430:
- [X] All tests pass
- [X] I have run `npm run doc`

### Description

@cypress/request is a fork of [request](https://github.com/request/request) that they use [Cypress](https://github.com/cypress-io/cypress). And is actually updated and also fix the problem of the vulnerability of tough-cookie dependency.

- [X] Change request to @cypress/request
- [X] Change request-promise to @cypress/request-promise

This change is not a breaking change and also can be safe for update in the current version without need to drop the support for old node versions until we do a more deep refactor of the package.

Related issues:

- https://github.com/yagop/node-telegram-bot-api/issues/1076
- https://github.com/yagop/node-telegram-bot-api/issues/1111
- https://github.com/yagop/node-telegram-bot-api/issues/1038

What is your opinion @yagop @kamikazechaser ?


saeedhei:
Does this module have the ability to update with Axios?

kamikazechaser (approved):
Defintely better to use a maintained fork.

yagop (approved):
LGTM! Clean solution

Issue #1012, opened 2022-10-01
Be able to use reactions on messages


fabianaferreira:
> Feature Request

Hello! It's my first time here, so hopefully, I'll do it correctly! I have searched for the implementation of the [sendReaction](https://core.telegram.org/method/messages.sendReaction) method from Telegram API on this Node implementation, but couldn't find anything.

### Introduction
I have a bot that gets kind of verbose due to the replies that it does whenever handling some command. Therefore, it would be handy to have a way of reacting to the messages instead of replying to them. It seems to be a good first issue for me to work on, given that it should be pretty similar to the `sendMessage` method implementation.

### Example
How `sendMessage` is implemented:
```javascript
  /**
   * Send text message.
   * @param  {Number|String} chatId Unique identifier for the target chat or username of the target channel (in the format `@channelusername`)
   * @param  {String} text Text of the message to be sent
   * @param  {Object} [options] Additional Telegram query options
   * @return {Promise} On success, the sent [Message](https://core.telegram.org/bots/api#message) object is returned
   * @see https://core.telegram.org/bots/api#sendmessage
   */
  sendMessage(chatId, text, form = {}) {
    form.chat_id = chatId;
    form.text = text;
    return this._request('sendMessage', { form });
  }
```

How I think `sendReaction` could be implemented:
```javascript
sendReaction(chatId, messageId, reaction, form = {}) {
    form.chat_id = chatId;
    form.reaction = reaction;
    form.msg_id = messageId;
    return this._request('sendReaction', { form });
}
```

Of course, this is just the high-level implementation; I still need to go through the `_request` method and add the necessary types and function signatures to the typings package.

danielperez9430:
The think is the Telegram Bot API don't allow you to send reactions or animate emojis in your messages, actually is no possible because the Telegram backend don't allow to do it with the Bot API. Unless Telegram support it, we can't do nothing with it.

fabianaferreira:
Hello, @danielperez9430! How are you? 

Thank you so much for clarifying it to me, honestly, I didn’t see that the link was for the client APi. Well, with that, I’m pretty sure that I can close the issue as a won’t do, since it’s not possible. Thanks one more time!

ponez:
@danielperez9430  what about this one ? 

https://core.telegram.org/methods#working-with-message-reactions

danielperez9430:
> @danielperez9430  what about this one ? 
> 
> 
> 
[...]

PR #1170, opened 2024-02-01
Added setMessageReaction

fede1132:
- [x] All tests pass

### Description

Added implementation of setMessageReaction telegram api

### References

Issue: https://github.com/yagop/node-telegram-bot-api/issues/1012

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -39,6 +39,7 @@ TelegramBot
         * [.logOut([options])](#TelegramBot+logOut) ⇒ <code>Promise</code>
         * [.close([options])](#TelegramBot+close) ⇒ <code>Promise</code>
         * [.sendMessage(chatId, text, [options])](#TelegramBot+sendMessage) ⇒ <code>Promise</code>
+        * [.setMessageReaction(chatId, messageId, [options])](#TelegramBot+setMessageReaction) ⇒ <code>Promise</code>
         * [.forwardMessage(chatId, fromChatId, messageId, [options])](#TelegramBot+forwardMessage) ⇒ <code>Promise</code>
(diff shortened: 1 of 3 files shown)


kamikazechaser (approved):
LGTM

RichardPickman:
> We have this method already from other more old pull request. Thank you also for your contribution ✋
> 
> Note: It is not necessary to use "JSON.stringify", in the library we use the "stringify" helper function

Is there any info how it is called now? Can't find any

danielperez9430:
> > We have this method already from other more old pull request. Thank you also for your contribution ✋
> > Note: It is not necessary to use "JSON.stringify", in the library we use the "stringify" helper function
> 
> Is there any info how it is called now? Can't find any

Is the method `setMessageReaction` 

Include since the 0.65 version: 

https://github.com/yagop/node-telegram-bot-api/blob/master/CHANGELOG.md#06500650---2024-02-20

RichardPickman:
> > > We have this method already from other more old pull request. Thank you also for your contribution ✋
> > > Note: It is not necessary to use "JSON.stringify", in the library we use the "stringify" helper function
> > 
> > 
> > Is there any info how it is called now? Can't find any
> 
> Is the method `setMessageReaction`
> 
> Include since the 0.65 version:
> 
> https://github.com/yagop/node-telegram-bot-api/blob/master/CHANGELOG.md#06500650---2024-02-20
> 
> Use example:
> 
> bot.setMessageReaction(chatId, messageId, options)
> 
> Note: options will be an object with the optional params that you can see in the telegram bot api docs.
> image Reference: https://core.telegram.org/bots/api#setmessagereaction

OOOOOH, I GET IT NOW. The version of types package is ^0.64.7, that is the reason why I don't get any suggestions with types. But the package itself is version 0.66

Issue #1174, opened 2024-02-21
Stop listening for new message events


wagnerfrana94:
Aggregate error causes the app to stop listening for new message events

error: [polling_error] {"code":"EFATAL","message":"EFATAL: AggregateError"}
error: [polling_error] {"code":"EFATAL","message":"EFATAL: Error: read ETIMEDOUT"}

Can someone help me?

yagop:
`EFATAL: AggregateError` and `read ETIMEDOUT` are network errors - the process couldn't reach `api.telegram.org` (DNS, firewall/egress, a flaky connection, or Telegram briefly unreachable), not a bug in your bot logic.

On the current TypeScript rewrite (**v1.1.0**) the polling loop is resilient to these: a failed `getUpdates` emits the `polling_error` event and the loop automatically schedules the next poll, so a transient network error no longer stops the bot from listening or crashes it. Just handle the event so failures are visible:

```js
bot.on('polling_error', (err) => console.error('polling_error', err.code, err.message));
```

If the host has chronic connectivity problems to Telegram, address it at the network level (DNS / egress / proxy). Closing - the "stops listening / crash" behavior is handled by the rewrite's loop, and the underlying `ETIMEDOUT` is environmental.

PR #1257, opened 2025-02-08
RFC: fix: FatalError set reference to causing error

ivanjh:
- [ ] All tests pass _(no - I don't have suitable tokens to test all integration tests)_
- [ ] I have run `npm run doc` _(not part of external api)_

### Description
FatalError does not provide enough context for understanding underlying issues.
Where available, attach the "causing" error to the FatalError to allow later inspection on `.cause` (see https://nodejs.org/api/errors.html#errorcause)

Serializing Error objects with `util.inspect()` recursively serializes `error.cause` if it is set and would allow better log messages in consuming applications.

I don't believe the existing approach of copying of the stack from the causing error is "least surprise" for consuming authors, but I do not yet understand reasons for why it is as is, or likelihood of breaking existing code.

### References
Recent node versions introduced a bug (https://github.com/nodejs/node/issues/54359) in certain network configurations that is becoming more numerous triggering EFATAL errors.

Reports often miss much depth, with only "EFATAL: AggregateError" as a description:
* Issues at this repo: https://github.com/yagop/node-telegram-bot-api/issues/1174 
* Issue on consuming repo: https://github.com/windkh/node-red-contrib-telegrambot/issues/345

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -30,7 +30,10 @@ exports.FatalError = class FatalError extends exports.BaseError {
     const error = (typeof data === 'string') ? null : data;
     const message = error ? error.message : data;
     super('EFATAL', message);
-    if (error) this.stack = error.stack;
+    if (error) {
+      this.stack = error.stack;
+      this.cause = error;
+    }
   }
 };
 


windkh:
+1

tedtribe:
+1

danielperez9430:
LGTM

Vacxe:
Please prepare a new release

Issue #1266, opened 2025-05-07
How to get reactions to messages?


autoantwort:
I want to get reactions to my messages and use the following code:
```js
const TelegramBot = require('node-telegram-bot-api');
const TOKEN = '***';

const bot = new TelegramBot(TOKEN, {
    polling: {
        params: {
            allowed_updates: ["message", "message_reaction"],
        }
    }
});

console.log(TelegramBot.messageTypes)

bot.on('message', (msg) => {
    bot.sendMessage(msg.chat.id, 'Try reacting to this message!');
});

bot.on('message_reaction', (reaction) => {
    console.log('Reaction received (message_reaction):', reaction);
});
bot.on("message_reaction_count", async (reaction) => {
    console.log("message_reaction_count", reaction);
});
console.log('Bot started with polling enabled');
```
But when i send the bot a message and react to it I don't get a update. Do you have a idea what the problem could be? 
<details><summary>Console output</summary>
<p>

```console
✗ DEBUG=* bun run ./test.js
  node-telegram-bot-api polling with options: {"allowed_updates":["message","message_reaction"],"offset":0,"timeout":10} +0ms
  node-telegram-bot-api HTTP request: {"form":{"allowed_updates":["message","message_reaction"],"offset":0,"timeout":10},"method":"POST","url":"https://api.telegram.org/bot***/getUpdates","simple":false,"resolveWithFullResponse":true,"forever":true} +0ms
[
  "text", "animation", "audio", "channel_chat_created", "contact", "delete_chat_photo", "dice",
  "document", "game", "group_chat_created", "invoice", "left_chat_member", "location", "migrate_from_chat_id",
  "migrate_to_chat_id", "new_chat_members", "new_chat_photo", "new_chat_title", "passport_data",
  "photo", "pinned_message", "poll", "sticker", "successful_payment", "supergroup_chat_created",
  "video", "video_note", "voice", "video_chat_started", "video_chat_ended", "video_chat_participants_invited",
  "video_chat_scheduled", "message_auto_delete_timer_changed", "chat_invite_link", "chat_member_updated",
  "web_app_data", "message_reaction"
]
Bot started with polling enabled
[...]

PR #1270, opened 2025-07-02
feat: stringify allowed_updates if it's an array on getUpdates

alfanzain:
- [ ] All tests pass - _I've tried to run those, but they even failed in `master`_

### Description

The bot was not receiving updates for `message_reaction` events, even though `allowed_updates` was specified in the polling config.
Turns out, the `allowed_updates` parameter was not being correctly passed to the getUpdates call if it is an array. It needs to be JSON-stringified.

This PR fixes the polling configuration to ensure that `allowed_updates` is properly passed as an array to the getUpdates method.

### References

* Issue #1266
* Telegram Bot API - Getting updates: https://core.telegram.org/bots/api#getupdates

### DIFF
--- a/src/telegram.js
+++ b/src/telegram.js
@@ -130,6 +130,8 @@ class TelegramBot extends EventEmitter {
    * @param {Object} [options.polling.params] Parameters to be used in polling API requests.
    *  See https://core.telegram.org/bots/api#getupdates for more information.
    * @param {Number} [options.polling.params.timeout=10] Timeout in seconds for long polling.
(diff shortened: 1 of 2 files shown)


src/telegram.js
danielperez9430, line 1: if (form.allowed_updates)) {
      form.allowed_updates = stringify(form.allowed_updates);
    }

GochoMugo (approved):
@alfanzain Confirmed manually that this fix works. Without stringifying `allowed_updates`, the parameter takes no effect in the API. Thanks!

PR #1278, opened 2025-10-03
update depencies + all typescript

el-j:
image

this pr moves to typescript, works with node20+ and has no vulnerabilities in build dependencies.


california444:
+1, please merge

PR #1299, opened 2026-04-07
Security: Webhook authentication uses substring match on URL token

tuanaiseo:
## Problem

Webhook authorization accepts any request where `req.url.indexOf(this.bot.token) !== -1`. This is not an exact route check and can be satisfied by crafted paths/query strings containing the token anywhere, increasing risk of unintended endpoint acceptance if the token is exposed.

**Severity**: `high`
**File**: `src/telegramWebHook.js`

## Solution

Match the webhook path exactly (for example, strict equality against `/bot${token}` or a fixed configured path), and additionally validate Telegram's `X-Telegram-Bot-Api-Secret-Token` header.

## Changes

- `src/telegramWebHook.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/telegramWebHook.js
+++ b/src/telegramWebHook.js
@@ -133,7 +133,13 @@ class TelegramBotWebHook {
     debug('WebHook request URL: %s', req.url);
     debug('WebHook request headers: %j', req.headers);
 
-    if (req.url.indexOf(this.bot.token) !== -1) {
+    const urlPath = req.url.split('?')[0];
+    const expectedPath = `/bot${this.bot.token}`;
+    const hasValidPath = urlPath === expectedPath;
+    const hasValidSecretToken = !this.options.secretToken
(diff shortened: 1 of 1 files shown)


yagop:
Thanks for the report. This patch targets `src/telegramWebHook.js`, which no longer exists after the v1.0.0 TypeScript rewrite — the webhook server is now `src/webhook.ts`, so it can't be merged as-is.

On the underlying concern: v1.0.0 already hardens this path. The handler strips the query string before matching the token (so it can't be smuggled via `?...=token`), and it adds support for Telegram's `X-Telegram-Bot-Api-Secret-Token` through the `secretToken` webhook option — the recommended way to authenticate webhook calls. When `secretToken` is set, every request must present the exact matching header or it's rejected with 401.

I'd recommend enabling `secretToken` for webhook deployments. Closing this PR as obsolete against the current codebase; thanks again for flagging it.

PR #1300, opened 2026-04-07
Security: Webhook request body is buffered without size limit (DoS risk)

tuanaiseo:
## Problem

Incoming webhook bodies are fully buffered with `bl` before parsing JSON, with no explicit maximum size. An attacker can send very large POST bodies to exhaust memory and degrade or crash the process.

**Severity**: `high`
**File**: `src/telegramWebHook.js`

## Solution

Enforce a strict request size limit before buffering (for example via `Content-Length` checks and stream byte counting), reject oversized payloads with 413, and set server/request timeouts.

## Changes

- `src/telegramWebHook.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/telegramWebHook.js
+++ b/src/telegramWebHook.js
@@ -3,7 +3,6 @@ const debug = require('debug')('node-telegram-bot-api');
 const https = require('https');
 const http = require('http');
 const fs = require('fs');
-const bl = require('bl');
 
 class TelegramBotWebHook {
   /**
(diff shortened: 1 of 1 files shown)


yagop:
Thanks for the report. This patch targets `src/telegramWebHook.js`, which no longer exists after the v1.0.0 TypeScript rewrite. The concern is already addressed in the current `src/webhook.ts`: the request reader enforces a hard payload cap (`MAX_PAYLOAD_BYTES`, 50MB) — it counts bytes as they arrive and calls `req.destroy()` the moment the limit is exceeded, rejecting the request rather than buffering it unboundedly, so oversized bodies can't exhaust memory.

Closing as already fixed in v1.0.0; thanks for raising it.

PR #1312, opened 2026-06-13
fix(webhook): authorize the token as a full path segment, not a substring

yagop:
### Problem

The webhook handler authorized any request whose path **contained** the bot token as a substring (`pathname.includes(this.bot.token)`). That's looser than an exact route check — a crafted path that merely embeds the token anywhere in a segment would authorize. Raised in #1299.

### Change

`src/webhook.ts` now matches the token as a **complete `/`-delimited path segment** (`_isAuthorizedPath`):

- ✅ Still authorizes: the documented `/bot<token>` path, a bare `/<token>`, or either nested under a custom prefix — so existing deployments are unaffected.
- 🔒 Now rejects with `401`: paths that only embed the token, e.g. `/bot<token>extra` or `/prefix<token>`.

`secretToken` (Telegram's `X-Telegram-Bot-Api-Secret-Token`) remains the stronger, recommended authentication and is unchanged.

### Tests

`test/unit/webhook.test.ts` — added two cases: a bare `/<token>` segment is accepted (compat), and a token-embedding-but-not-segment path is rejected with 401 (the hardening).

- `npm run typecheck` — clean
- webhook unit tests — 17 pass / 0 fail
- full unit suite — 93 pass / 0 fail

### Notes

- Behavior change documented under `[Unreleased] → Security` in `CHANGELOG.md`.
- Compatibility: the only setups affected are those whose webhook path embeds the token *within* a larger segment (not as `/bot<token>` or `/<token>`), which is unusual.

### References

- Addresses #1299

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][Unreleased]
 
+### Security
+
+- **Webhook path authorization now matches the bot token as a complete path
+  segment** instead of a raw substring. A request is authorized only when the
+  token is a full `/`-delimited segment: the documented `/bot<token>` path, a
+  bare `/<token>`, or either nested under a custom prefix. Paths that merely
+  embed the token (e.g. `/bot<token>extra`) are now rejected with `401`. For
+  stronger authentication, set the `secretToken` option to also validate
+  Telegram's `X-Telegram-Bot-Api-Secret-Token` header.
+
 ## [1.0.0][1.0.0] - 2026-06-12
 
 ### Rewritten in TypeScript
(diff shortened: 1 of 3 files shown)


yagop:
On reflection, I'm closing this without merging.

The bot token *is* the shared secret here: the webhook is authenticated by the caller presenting the correct token in the path. Under that model, any request carrying the right token is legitimate by definition, and a request that doesn't know the token can't authorize regardless of whether we match it as a substring or as a full path segment. So tightening `includes(token)` to exact-segment matching doesn't change the threat model — it only risks breaking unusual-but-valid path layouts that the substring check accepted.

For anyone wanting an additional, header-based check, the `secretToken` option (Telegram's `X-Telegram-Bot-Api-Secret-Token`) remains available and is the recommended defense-in-depth.

Closing as won't-merge; the current substring behavior is fine. Thanks for the discussion on #1299.

Issue #1189, opened 2024-04-07
The editMessageMedia method is working a bit wrong


Toper9636:
### Expected Behavior
I expected to be able to change the media in the message using Buffer or Stream

### Actual Behavior
I can change media using only file_id/path/url

### Steps to reproduce the Behavior
Use  `TelegramBot#editMessageMedia`

### In other libraries

I looked in Telegraf and there you can change media using Buffer or Stream

Sp3rick:
that's true af
i found a fix i will try to pull it soon

yagop:
Confirmed and fixed. `editMessageMedia` now accepts a **Buffer / stream / local path** for the new `media` (and its `thumbnail` / `cover`), uploading via an `attach://` part - it previously only handled `attach://<local-path>` / file_id / URL.

Fix in PR #1315 (verified against the live API). It's non-breaking - string callers and the old `attach://<local-path>` form still work. Leaving this open until it's released.

```js
await bot.editMessageMedia(
  { type: 'photo', media: fs.readFileSync('new.png') },
  { chat_id, message_id },
);
```

yagop:
Fixed by #1315 and released in v1.1.1. Closing as completed.

PR #1315, opened 2026-06-14
fix: support Buffer/stream/path uploads in editMessageMedia (#1189)

yagop:
## Problem

`editMessageMedia` could only upload a *new* file when `media` was passed as `attach://<local-path>`. A **Buffer / stream**, or even a plain local path, was unsupported (the field was typed `string`), and the secondary fields (`thumbnail` / `cover`) were string-only. Every other send method accepts a Buffer/stream/path, so this was an inconsistency.

Refs #1189

## Fix

Route `media` through the same `_buildMediaItems` pipeline used by `sendMediaGroup` / `sendPaidMedia`:

- The primary `media` and its `thumbnail` / `cover` now accept a **Buffer / stream / local path** (uploaded as an `attach://` multipart part) or a **file_id / URL** string (passed through).
- `media` is widened from `InputMedia` to `Uploadable<InputMedia>` (the existing helper that widens file fields to `InputFile`).
- `_buildMediaItems`'s parameter type is broadened to also accept `Uploadable<InputMedia>`.

```js
// Now works (previously a type error / unsupported):
await bot.editMessageMedia(
  { type: 'photo', media: fs.readFileSync('new.png') },
  { chat_id, message_id },
);
```

## Non-breaking

- `string` callers are unaffected (`string` is assignable to `InputFile`).
- The legacy `attach://<local-path>` form is still accepted (the prefix is stripped and the path uploaded).
- `file_id` / URL callers still produce a urlencoded body (no multipart).

## Tests / verification

- `npm run typecheck` clean; `npm run build` clean.
- Unit (`test/unit/telegram.test.ts`): **45/45 pass**, incl. 3 new (Buffer -> `attach://0_media`; video `thumbnail` Buffer -> `attach://0_thumbnail`; file_id -> urlencoded passthrough).
- **Live API**: `editMessageMedia` scoped run = **3/3 pass** (attach:// path, **Buffer upload**, file_id with caption + reply_markup).
- Regenerated `doc/api.md`; corrected a now-stale `sendMediaGroup` doc comment that claimed editMessageMedia's secondary fields were string-only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   });
   ```
 
+### Fixed
+
+- `editMessageMedia` now accepts a Buffer / stream / local file path for the new
+  `media` (and its `thumbnail` / `cover`), uploading it via an `attach://` part -
+  previously only a file_id / URL or the legacy `attach://<local-path>` form
+  worked. Resolved through the same `_buildMediaItems` pipeline as
+  `sendMediaGroup`; string callers and the old `attach://<local-path>` form are
+  unaffected.
+  ([#1189](https://github.com/yagop/node-telegram-bot-api/issues/1189))
+
 ## [1.1.0][1.1.0] - 2026-06-13
 
 ### Bot API 10.1 (June 11, 2026)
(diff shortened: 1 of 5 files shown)


danielperez9430:
LGTM

PR #1323, opened 2026-06-23
chore: add CommonJS export and main; bump version to 1.1.2

### LINKED ISSUE #1322: v1 ESM-only migration breaks CJS consumers (NestJS, Jest, etc.)
### Description

Upgrading to v1 breaks any project that consumes this package in a CommonJS context. Node.js throws `ERR_PACKAGE_PATH_NOT_EXPORTED` at runtime because the new `exports` map has no `require` condition.

Most notably, **NestJS — one of the most popular Node.js frameworks — compiles to CJS by default and cannot use this package at v1 at all.** The same applies to any project using Jest in its default configuration.

### Environment

- `node-telegram-bot-api`: `1.1.1`
- `Node.js`: `v22.14.0`
- Framework: NestJS (compiles to CJS by default)

### Error

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in
.../node_modules/node-telegram-bot-api/package.json
    code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
```

### Root cause

The current `package.json` in v1.1.1 defines only an `import` condition in the `exports` map, with no `require` condition:

```json
{
  "name": "node-telegram-bot-api",
  "version": "1.1.1",
  "type": "module",
  "main": "./dist/index.js",
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.js"
    }
  }
}
```

Although `main` is set, Node.js v22 **ignores `main` when an `exports` field is present** and strictly resolves only the declared conditions. Since there is no `"require"` condition, any `require('node-telegram-bot-api')` call — including those emitted by TypeScript/NestJS builds — throws at runtime.

CJS/ESM interop is also **asymmetric by design**: ESM can import CJS, but CJS cannot import ESM. This means NestJS users have no workaround short of pinning to v0.x.

### Impact

This is a **breaking change with no migration path** for a significant portion of the ecosystem:

- **NestJS** compiles to CJS by default and [has an open ESM migration issue since 2021](https://github.com/nestjs/nest/issues/7021) with no committed timeline. All NestJS users are blocked from upgrading to v1.
- **Jest** runs in CJS mode unless explicitly configured with `--experimental-vm-modules`. Most projects don't do this.
- Any TypeScript project using `"module": "CommonJS"` in `tsconfig.json` — which is still the recommended default for Node.js apps — is affected.

### Suggested fix

Add a `require` condition pointing to a CJS build, which is the dual-package approach recommended by the Node.js docs:

```json
{
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/index.mjs",
      "require": "./dist/index.cjs"
    }
  }
}
```

This is a non-breaking approach and is how most major packages (`chalk`, `got`, etc.) have handled the same migration.

### Workaround (for now)

Pin to the last v0.x release:

```bash
npm install node-telegram-bot-api@0.66.0
```

parirox:
- [x] All unit tests pass (`npm run test:node:unit`)
- [x] All integration tests pass (`npm run test:node:integration`)
- [x] `npm run typecheck` is clean
- [ ] `doc/api.md` is up to date (`bun run generate:docs` leaves no diff — required when a `TelegramBot` method signature changed)

### Description

Adds a CommonJS build entry (`./dist/index.cjs`) and a `require` condition to the `exports` field, so the package works as a dual ESM/CJS module. Also sets `main` to the CJS build for legacy tooling compatibility and bumps the version to `1.1.2`.

This unblocks NestJS users and any other CJS consumer (e.g. Jest with default config) from upgrading to v1, who currently hit `ERR_PACKAGE_PATH_NOT_EXPORTED` at runtime.

### References

* Closes #1322

---
### Running the test suite
...

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,9 @@ Node.js module to interact with the official [Telegram Bot API](https://core.tel
 npm i node-telegram-bot-api
 ```
 
-> ✍️ **Note:** This package is **ESM-only** and requires **Node.js ≥ 18**. Use
-> `import` (not `require`). **TypeScript types are bundled** — do **not** install
+> ✍️ **Note:** This package provides both **ESM** and **CommonJS (CJS)** builds and
+> works with **Node.js ≥ 18**. Use `import` in ESM environments and `require` in
+> CommonJS environments. **TypeScript types are bundled** — do **not** install
 > `@types/node-telegram-bot-api` (it would shadow the bundled, more accurate
 > types). Upgrading from `0.6x`? See the [migration notes in the changelog][migration].
 
@@ -57,6 +58,20 @@ bot.on('message', (msg) => {
 
   // send a message to the chat acknowledging receipt of their message
   bot.sendMessage(chatId, 'Received your message');
+});
+```
+
+```js
+// CommonJS (require)
+const TelegramBot = require('node-telegram-bot-api');
+
+const token = 'YOUR_TELEGRAM_BOT_TOKEN';
+
+const botCjs = new TelegramBot(token, { polling: true });
+
+botCjs.on('message', (msg) => {
+  const chatId = msg.chat.id;
+  botCjs.sendMessage(chatId, 'Received your message');
 });
 ```
 
--- a/package.json
+++ b/package.json
@@ -1,14 +1,15 @@
 {
   "name": "node-telegram-bot-api",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "description": "Telegram Bot API — modern TypeScript rewrite with generated types",
   "type": "module",
-  "main": "./dist/index.js",
+  "main": "./dist/index.cjs",
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/index.js"
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
     }
   },
   "files": [


yagop:
Thanks for tackling #1322. The current change is not enough yet because it only advertises a CommonJS entry, but does not actually build one.

`package.json` now points both `main` and the `exports.require` condition at `./dist/index.cjs`, but the build script is still just `tsc -p tsconfig.build.json`. With the current `tsconfig` (`module: NodeNext`, root package `type: module`), that emits `dist/index.js` and declarations, not `dist/index.cjs`.

I verified this on the PR branch:

```sh
npm ci
npm pack --dry-run
node -e "require('node-telegram-bot-api')"
```

The packed files include `dist/index.js`, but no `dist/index.cjs`, and the `require()` check fails with:

```text
Cannot find module '/tmp/ntba-pr1323/dist/index.cjs'
```

So CommonJS users would still be broken, just with `MODULE_NOT_FOUND` instead of `ERR_PACKAGE_PATH_NOT_EXPORTED`.

The fix needs to generate a real CJS artifact and test it. Two reasonable options:

1. Use a dual-package build tool such as `zshy`, which emits both `index.js` and `index.cjs`.
2. Keep `tsc` and add a second CJS build, for example to `dist/cjs`, with a nested `package.json` containing `{ "type": "commonjs" }`, then point `exports.require` at that file.

If we are okay with CommonJS consumers using either of these shapes:

```js
const { TelegramBot } = require('node-telegram-bot-api');
// or
const TelegramBot = require('node-telegram-bot-api').default;
```

then no compatibility wrapper is needed. But the README example in this PR should use one of those forms instead of `const TelegramBot = require(...)`.

Also, since the PR bumps `package.json` to `1.1.2`, `package-lock.json` should be updated too; it still reports `1.1.1` for the root package.

yagop:
Closed by https://github.com/yagop/node-telegram-bot-api/pull/1325

PR #1325, opened 2026-06-24
[codex] build: add zshy dual module output

### LINKED ISSUE #1322: v1 ESM-only migration breaks CJS consumers (NestJS, Jest, etc.)

yagop:
## Summary

- Switch the v1 build to zshy so the package emits both ESM and CommonJS artifacts from the TypeScript entrypoint.
- Add conditional `exports` entries for ESM and CJS, with CJS pointing at `dist/index.cjs` and ESM at `dist/index.js`.
- Keep the supported CommonJS usage as destructuring: `const { TelegramBot } = require('node-telegram-bot-api');`.
- Bump the package version and lockfile metadata to `1.1.2`.

Closes #1322.

## Validation

- `npm run build`
- `node -e "const { TelegramBot } = require('node-telegram-bot-api'); if (typeof TelegramBot !== 'function') process.exit(1);"`
- `node --input-type=module -e "import TelegramBot, { TelegramBot as NamedTelegramBot } from 'node-telegram-bot-api'; if (typeof TelegramBot !== 'function' || typeof NamedTelegramBot !== 'function') process.exit(1);"`
- CommonJS TypeScript temp consumer compile with `module: CommonJS`
- ESM TypeScript temp consumer compile with `module: NodeNext`
- `npm run typecheck`
- `npm run test:node:unit`
- `npm pack --dry-run`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,28 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][Unreleased]
 
+## [1.1.2][1.1.2] - 2026-06-24
+
+### Added
+
+- **CommonJS consumption restored.** The package now ships a dual ESM + CJS
+  build (produced by `zshy`): the `package.json` `exports` map exposes both an
+  `import` and a `require` condition (with their own `.d.ts` / `.d.cts` typings),
+  so the library can be loaded with either syntax —
+
+  ```js
+  // CommonJS
+  const { TelegramBot } = require("node-telegram-bot-api");
+
+  // ESM (unchanged)
+  import TelegramBot from "node-telegram-bot-api";
+  ```
+
+  The v1.0.0 dynamic-import workaround
+  (`const { default: TelegramBot } = await import("node-telegram-bot-api")`) is
+  no longer required. Source maps are emitted for both module formats, and each
+  CJS artifact references its own map.
+
 ## [1.1.1][1.1.1] - 2026-06-22
 
 ### Added
@@ -817,4 +839,5 @@ Fixed:
 [1.0.0]:https://github.com/yagop/node-telegram-bot-api/releases/tag/v1.0.0
 [1.1.0]:https://github.com/yagop/node-telegram-bot-api/releases/tag/v1.1.0
 [1.1.1]:https://github.com/yagop/node-telegram-bot-api/releases/tag/v1.1.1
-[Unreleased]:https://github.com/yagop/node-telegram-bot-api/compare/v1.1.1...master
+[1.1.2]:https://github.com/yagop/node-telegram-bot-api/releases/tag/v1.1.2
+[Unreleased]:https://github.com/yagop/node-telegram-bot-api/compare/v1.1.2...master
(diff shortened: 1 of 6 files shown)


yagop:
The CJS source-map finding from the review turned out to be an upstream zshy bug, not something in this PR's config. zshy renames the emitted output (.js→.cjs, .d.ts→.d.cts, and the map files) but writes the file contents unchanged, so the //# sourceMappingURL= comment and each map's "file" field keep pointing at the ESM names. It even reproduces in zshy's own published artifacts.

https://github.com/colinhacks/zshy/issues/71

PR #1329, opened 2026-07-01
v2: unhandled handler error kills the polling loop (add a default error boundary) (#1328)

### LINKED ISSUE #1328: v2: unhandled handler error kills the polling loop (add a default error boundary)
## Problem

In v2 (`2.0.0-alpha.1`), a bot that never installs `bot.catch()` dies on the first error thrown by any handler - and can enter a crash loop on restart.

The chain:

1. **`handleUpdate` rethrows.** With no error boundary installed, a middleware/handler throw escapes `handleUpdate` (`src/core/bot.ts`):

   ```ts
   } catch (err) {
     if (this.errorHandler) await this.errorHandler(err, ctx);
     else throw err; // <- no catch() installed: rethrow
   }
   ```

2. **The polling pump has no protection of its own.** `startPolling`'s `for await` body awaits `handleUpdate` with no try/catch, so the throw exits the loop, finalizes the `longPoll` generator, and rejects `startPolling()`'s promise. Polling is permanently stopped. The retry machinery in `longPoll`/`Transport` does not help: it only guards *fetching* updates, not *processing* them.

3. **The process usually dies too.** The common calling pattern is fire-and-forget (`bot.startPolling()` without an `await`/`.catch`), so the rejection is an unhandled promise rejection - fatal in Node by default.

4. **Poison-update crash loop.** `longPoll` advances the offset only after an update is handled, and Telegram only confirms updates once a later `getUpdates` call carries the advanced offset. The update that killed the process was never confirmed, so on restart Telegram redelivers it. If it deterministically triggers the same throw: start, receive same update, throw, die, repeat.

## Inconsistency between entry points

The codebase already takes a position on this - but only in one of the three modes:

- **Webhook + fastAck** (`src/core/webhook.ts`): handler rejections are caught, logged, and dropped, precisely so one bad update cannot do damage.
- **Webhook without fastAck**: the throw escapes the callback; whatever the runtime does with an unhandled rejection (Bun.serve, Workers, node:http adapters all differ) decides the response, typically a 500 -> Telegram redelivers.
- **Long polling**: the throw kills the loop and (usually) the process.

Same `handleUpdate`, three different failure behaviors depending on deployment mode.

## Proposed solution: a default error boundary

Give `errorHandler` a default of log-and-continue; `bot.catch()` replaces it:

```ts
private errorHandler: (err: unknown, ctx: Context) => unknown = (err, ctx) => {
  console.error(`Unhandled error while processing update ${ctx.update.update_id}`, err);
};
```

This is the Express model: an unhandled route error logs and 500s that one request; it never kills the server. A bot is a server - one malformed update must not take down service for every other chat.

Properties:

- Fixes the polling crash and the restart crash loop (the offset advances normally, so the poison update is confirmed and consumed).
- Unifies all three modes: *an unhandled handler error is logged and the update is consumed; install `bot.catch()` to change that.*
- Fail-loud stays one line away and remains explicit: `bot.catch((err) => { throw err; })` - a rethrow from the boundary still propagates.
- `console.error` is web-standard, so `src/core` stays edge-safe (no Node globals).

### Known semantic change

In non-fastAck webhook mode today, an escaping error typically produces a 500 and Telegram redelivers the update - a free (if crude) retry mechanism. With a default boundary the update is ACKed instead. This is arguably correct (redelivering an update that deterministically throws just hammers the endpoint, and fastAck mode already cannot un-ACK), but it must be called out in the docs.

## Scope

- [ ] Default `errorHandler` (log-and-continue) on `Bot`; `bot.catch()` replaces it; a throw from the user's boundary still propagates.
- [ ] Non-fastAck webhook path: catch the (now only user-rethrown) error and return an explicit 500 instead of relying on per-runtime unhandled-rejection behavior.
- [ ] Unit tests: polling survives a throwing handler; offset advances past the failing update; rethrow from `bot.catch()` still stops the loop; webhook returns 500 on boundary rethrow.
- [ ] Docs: JSDoc on `catch()`/`startPolling()` + README error-handling section spelling out the contract, including the webhook ACK semantics change.

yagop:
Closes #1328

- [x] T1  Bot: default log-and-continue error boundary; `bot.catch()` replaces it, a throw from the boundary still propagates (+ JSDoc + unit tests)
- [x] T2  webhookCallback: catch a boundary rethrow in the non-fastAck path and return an explicit 500 (+ unit tests)
- [x] T3  Docs: README error-handling section + regenerate doc/api.md


yagop:
😺 Michi — merged, all done 🐾

PR #1331, opened 2026-07-10
fix(v2): stream multipart file uploads

yagop:
- [x] `npm run check` is clean (typecheck src + test + examples, `lint:core`, `check:edge`, unit suite)
- [x] `npm run build` is clean
- [x] Generated files are untouched / regenerated (`doc/api.md` was regenerated; generated API sources are untouched)

### Description

Stream v2 multipart file uploads instead of converting every `ReadableStream` to a full in-memory `Blob` before calling `fetch`.

The old conversion made request bodies replayable, but a large upload temporarily required memory proportional to the whole file. This change adds a runtime-agnostic multipart stream encoder and passes the Web `ReadableStream` directly to `fetch`.

Retry behavior is explicit:

- A bare `ReadableStream` is one-shot and is not retried.
- An `InputFileStreamFactory` returns a fresh stream for each attempt and remains retryable.
- The Node-only `fromPath()` helper now uses a stream factory, so disk uploads are both bounded-memory and retryable.
- `Blob` and `Uint8Array` inputs keep using native `FormData`.

Node fetch receives `duplex: "half"` for streaming request bodies. The core remains free of Node imports and still passes the edge bundle check.

Validation:

- `npm run check` (112 Bun unit tests)
- `npm run test:node:unit` (112 Node unit tests)
- `npm run build`
- Real local HTTP uploads through Node fetch and Bun fetch
- Synthetic 512 MiB upload: about 2 MiB RSS growth while consuming the complete multipart body

### References

The failure was reproduced with a 731 MiB Telegram upload that exhausted the container while `inputFileToBlob()` called `new Response(stream).blob()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,11 @@ const bot = new Bot(process.env.BOT_TOKEN!);
 await bot.api.sendPhoto({ chat_id, photo: await fromPath("./cat.jpg") });
 // upload raw bytes (web-standard, runs anywhere)
 await bot.api.sendDocument({ chat_id, document: new InputFile(bytes, { filename: "report.pdf" }) });
+// stream a large file without buffering it; the factory makes transport retries possible
+await bot.api.sendVideo({
+  chat_id,
+  video: new InputFile(() => createVideoStream(), { filename: "video.mp4", contentType: "video/mp4" }),
+});
 
 // a raw InputFile nested in a structure is auto-hoisted to an attach:// part
 await bot.api.sendMediaGroup({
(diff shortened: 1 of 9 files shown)


yagop:
Superseded by #1332, which lands the same streaming multipart encoder and adopts this PR's InputFileStreamFactory and content-type hardening. Closing in its favor: validation there surfaced that Bun's fetch stalls forever on ReadableStream request bodies against HTTPS origins (local-HTTP uploads work, which is why this PR's validation passed), so streamed uploads - including every fromPath() upload here - would hang on Bun. #1332 adds a no-stream fallback (buffered replayable Blob) with Bun excluded from streaming until that is fixed upstream, and also removes the FormData double-copy for Blob/Uint8Array uploads.

PR #1332, opened 2026-07-10
feat(v2): stream multipart uploads instead of buffering them

yagop:
- [x] `npm run check` is clean (typecheck src + test + examples, `lint:core`, `check:edge`, unit suite: 130 pass on Node; on Bun 129 pass + 1 skip proxied, 130 pass direct)
- [x] `npm run build` is clean
- [x] Generated files regenerated (`doc/api.md`); generated API sources untouched

### Description

Uploads used to hold 2-4x the file size in RAM: `fromPath` read the whole file, `encodeForm` copied every part into `FormData` Blobs, and a `ReadableStream` `InputFile` was drained into an in-memory `Blob` before the send. This PR replaces the `FormData` encoder with a hand-rolled `multipart/form-data` body built as a web `ReadableStream` and passed straight to `fetch` (`duplex: "half"`), so file bytes flow from their source without ever being materialized. Measured on a 100MB upload: ~1MB retained vs 200-300MB before.

- **`src/core/multipart.ts`** - the streaming encoder. The body is laid out once as ordered pieces (encoded part headers + each file's `Blob`/`Uint8Array`/stream/factory); each send attempt streams the pieces into a fresh `ReadableStream`. Web-standard APIs only; `lint:core` + `check:edge` stay green.
- **Retry semantics are explicit**: `Blob`/`Uint8Array` uploads re-stream on retry (byte-identical); a bare `ReadableStream` is one-shot - sent once, a failure surfaces immediately; the new **`InputFileStreamFactory`** (`() => ReadableStream | Promise<ReadableStream>`) opens a fresh stream per attempt and stays retryable.
- **`fromPath()`** returns a disk-backed `Blob` via `fs.openAsBlob` (Node >= 19.8, feature-detected; older Nodes fall back to `readFile`), so disk uploads are flat-memory and retry-safe.
- **Runtimes that cannot stream a request body fall back** to sending the same bytes as one buffered, replayable `Blob`, chosen by a once-per-process probe. **Bun is excluded from streaming only when an HTTP(S) proxy is configured** (read via `Bun.env`, which Bun honors automatically): its `fetch` streams request bodies fine on a direct connection, but a stream body routed through a CONNECT proxy stalls forever (oven-sh/bun#33918; Node and curl pass through the same proxy, so no local probe can catch it). Direct-connection Bun streams like Node/Deno; drop the guard when the upstream issue is fixed.
- `meta.contentType` is guarded against CR/LF header injection into part headers; part names/filenames use the WHATWG escaping.
- `inputFileToBlob` is removed - nothing converts to `FormData` anymore.
- Default per-request `timeoutMs` raised 30s -> 5 min so large uploads are not cut off mid-stream (`getUpdates` already derives its own long-poll timeout).

### Validation

- Unit suites on Bun and Node, plus the full local gate.
- Live against api.telegram.org on **Node**: plain-bytes, `fromPath`, one-shot stream, stream factory, and nested `attach://` media-group uploads (chunked transfer-encoding accepted; confirmed independently with curl).
- Live against api.telegram.org on **Bun** via the buffered fallback, plus the scoped e2e upload methods.
- Synthetic 100MB upload memory measurement before/after (200-300MB retained -> ~1MB), including a full replay/retry re-send.

Supersedes #1331 (same goal, independently implemented): adopts its `InputFileStreamFactory` and content-type hardening, and additionally removes the `FormData` double-copy for `Blob`/`Uint8Array` uploads and adds the no-stream fallback that keeps proxied Bun working (streamed uploads there stall through CONNECT proxies, which local-HTTP validation does not catch).


yagop:
The Bun streaming exclusion now has an upstream anchor: filed oven-sh/bun#33918 (https://github.com/oven-sh/bun/issues/33918) with a dependency-free repro - `fetch()` with a `ReadableStream` request body stalls forever against HTTPS origins on Bun 1.3.14 (latest stable), while the same bytes as a plain body and the same stream against plain-HTTP origins both succeed; Node 26 passes all four combinations. The `supportsRequestStreams` guard comment references the issue; when it is fixed upstream, the guard comes off and Bun streams like Node/Deno.

yagop:
Upstream resolution: Bun triage confirmed oven-sh/bun#33918 and root-caused it to the CONNECT-tunnel stream-body path (fixed by oven-sh/bun#32635, first shipping in Bun 1.4.0; verified here on canary through the same proxy that stalled - all four combinations pass). The guard is now version-gated: only proxied Bun < 1.4.0 takes the buffered fallback, so the workaround retires itself as users upgrade. All four guard branches are unit-verified.

PR #1356, opened 2026-09-04
Transport: an empty-string apiRoot is used verbatim instead of falling back to the default (#1354)

### LINKED ISSUE #1354: Transport: an empty-string apiRoot is used verbatim instead of falling back to the default
## What happens

An empty-string `apiRoot` is not treated as "use the default"; it is used verbatim, producing malformed request URLs.

`Transport` resolves it with `??`, which only replaces `undefined`/`null`:

```ts
// src/core/transport.ts:106
this.apiRoot = (options.apiRoot ?? DEFAULT_API_ROOT).replace(/\/+$/, "");
// ...
// :128
const url = `${this.apiRoot}/bot${this.token}/${method}`;
```

So `new Api(token, { apiRoot: "" })` builds `/bot<token>/getMe` - a relative, invalid URL - and every request fails confusingly instead of hitting `https://api.telegram.org`.

## Why it bites

It is natural to wire `apiRoot` straight from an optional env var:

```ts
apiRoot: process.env.TEST_API_ROOT,   // "" when the var is set-but-empty
```

`TEST_API_ROOT=""` (or `TEST_API_ROOT=` in a shell / CI) then yields `apiRoot: ""`, not `undefined`, and the client silently points nowhere.

This is exactly the shape in `test/e2e/methods.test.ts`: the emulator toggle uses `Boolean(process.env.TEST_API_ROOT)` (empty string -> `false`, so it picks the *live* rate-limit profile) while `apiRoot: process.env.TEST_API_ROOT` passes the empty string straight through. With `TEST_API_ROOT=""` the two disagree and the whole suite talks to a relative URL.

## Repro

```ts
const api = new Api("123:abc", { apiRoot: "" });
await api.getMe(); // fetch on "/bot123:abc/getMe" - not api.telegram.org
```

## Suggested fix

Coalesce empty/whitespace to the default in `Transport`, so an env-driven `apiRoot` is safe:

```ts
const root = options.apiRoot?.trim();
this.apiRoot = (root ? root : DEFAULT_API_ROOT).replace(/\/+$/, "");
```

(Optionally validate that a non-empty value parses as an absolute `http(s)` URL and throw at construction otherwise, rather than failing per request.)

Separately, `test/e2e/methods.test.ts` should derive both the toggle and the option from one normalized value (`process.env.TEST_API_ROOT || undefined`) so `EMULATOR` and `apiRoot` cannot disagree.

## Notes

Found while adding the emulator wiring to the e2e suite (#1353).

yagop:
Closes #1354

- [x] T1  Transport: coalesce empty/whitespace `apiRoot` to the default so an env-driven value is safe, with a unit test
- [x] T2  e2e: derive both `EMULATOR` and `apiRoot` from one normalized `process.env.TEST_API_ROOT || undefined` so they cannot disagree

### DIFF
--- a/src/core/transport.ts
+++ b/src/core/transport.ts
@@ -103,7 +103,10 @@ export class Transport {
     options: TransportOptions = {},
   ) {
     if (!token) throw new TelegramBotError("A bot token is required", { code: "EPARAM" });
-    this.apiRoot = (options.apiRoot ?? DEFAULT_API_ROOT).replace(/\/+$/, "");
+    // Coalesce an empty/whitespace-only apiRoot (e.g. an unset-but-present env
+    // var) to the default; `??` alone would keep "" and build a relative URL.
+    const root = options.apiRoot?.trim();
+    this.apiRoot = (root ? root : DEFAULT_API_ROOT).replace(/\/+$/, "");
     const fetchImpl = options.fetch ?? globalThis.fetch;
     if (typeof fetchImpl !== "function") {
       throw new NetworkError("No fetch implementation available; pass options.fetch");
--- a/src/core/transport.ts
+++ b/src/core/transport.ts
@@ -103,8 +103,7 @@ export class Transport {
     options: TransportOptions = {},
   ) {
     if (!token) throw new TelegramBotError("A bot token is required", { code: "EPARAM" });
-    // Coalesce an empty/whitespace-only apiRoot (e.g. an unset-but-present env
-    // var) to the default; `??` alone would keep "" and build a relative URL.
+    // Empty/whitespace apiRoot falls back to the default; `??` would keep "".
     const root = options.apiRoot?.trim();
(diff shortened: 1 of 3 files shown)


yagop:

**Done**
- T1 (`217494c`): Transport coalesces empty/whitespace `apiRoot` to the default; added unit tests covering `""`, whitespace, and a non-empty root.
- T2 (`8326cb0`): e2e `methods.test.ts` derives `EMULATOR` and `apiRoot` from one normalized `TEST_API_ROOT?.trim() || undefined`, so they can't disagree.
- `486a1d7`: simplified the added comments.

**CI**: GitHub Actions (typecheck, build, unit on Bun + Node 22/24/26) and AppVeyor all pass; E2E is skipped by design.

Waiting for your merge. 🐾

yagop:
😺 Michi — merged, all done 🐾

PR #1358, opened 2026-09-05
longPoll: a 409 Conflict permanently stops the polling loop (#1350)

### LINKED ISSUE #1350: longPoll: a 409 Conflict permanently stops the polling loop
## What happens

A `409 Conflict` from `getUpdates` ("terminated by other getUpdates request; make sure that only one bot instance is running") kills the long-poll pump permanently.

`isTransientError` (`src/core/errors.ts:102-105`) counts only `NetworkError`, `TimeoutError`, 429 and 5xx as transient:

```ts
export function isTransientError(err: unknown): boolean {
  if (err instanceof NetworkError || err instanceof TimeoutError) return true;
  if (err instanceof TelegramApiError) return err.errorCode === HTTP_STATUS_TOO_MANY_REQUESTS || err.errorCode >= 500;
  return false;
}
```

so `longPoll` takes the `if (!retry || !isTransientError(err)) throw err` branch (`src/core/longpoll.ts:59`), `startPolling` rejects, and the loop is over. No retry, no backoff, no resumption.

## Why that is wrong

409 almost always means *another instance is still polling*, and that is usually temporary and self-healing:

- a redeploy where the old process has not exited yet;
- a supervisor restart overlapping by a second or two;
- a developer running the bot locally against the same token;
- any second `getUpdates` against the token, including a one-off `curl` while debugging.

Every comparable library treats it as recoverable - log it, back off, keep polling - because the competing poller normally goes away on its own. Here, a two-second overlap during a deploy takes the bot down until someone restarts it.

## Repro

With a bot polling via `run(bot)` / `bot.startPolling()`:

```sh
curl -s "https://api.telegram.org/bot$TOKEN/getUpdates?timeout=0&limit=1"
```

Telegram answers the *bot's* in-flight call with 409. The process stays alive (see #TODO for that half) and never polls again: no `getUpdates` goes out, `pending_update_count` climbs, and nothing is handled.

## Suggested fix

Treat 409 as transient so the loop backs off and resumes. Either add it to `isTransientError`, or - better, since 409 is not transient for a *request*, only for polling - handle it in `longPoll`:

```ts
const CONFLICT = 409;
const isPollConflict = (err: unknown) => err instanceof TelegramApiError && err.errorCode === CONFLICT;
// ...
if (!retry || !(isTransientError(err) || isPollConflict(err))) throw err;
```

A conflict-specific backoff (longer than the 1s default, since the other instance needs time to exit) and an explicit `log()` line would help operators see what is happening. `onError` already gives callers a hook to observe it.

Worth deciding: retry forever, or give up after N consecutive conflicts so a genuinely misconfigured two-instance deployment is not silently masked? A bounded retry that then throws seems the honest default.

## Notes

Found while running `examples/17-callback-tracking.ts` live: a single diagnostic `getUpdates` curl silently ended the bot's polling loop.

yagop:
Closes #1350

- [x] T1  `longPoll`: treat a 409 poll conflict as recoverable - back off (its own longer delay) and resume, bounded by a max consecutive-conflict count that then throws; log + `onError`, counter resets on any successful poll
- [x] T2  Unit tests: resume after a transient 409, throw after N consecutive conflicts, and respect `retry: false`
- [x] T3  Regenerate `doc/api.md` for the new `LongPollOptions` fields
- [x] T4  `run()`: a fatal poll-stop must not be silent - the managed runner surfaces it to stderr before re-throwing, so a dropped/late-observed rejection can't leave the process alive with dead polling; unit test
- [x] T5  `startWebhook`: cap graceful shutdown so an idle keep-alive (or stuck) connection can't leave a stopped-but-never-exiting process - drop idle connections, force-close past `shutdownTimeoutMs`; unit test + docs
- [x] T6  Bump the documented minimum to Node 18.2 (`engines`, docs) - `closeIdleConnections`/`closeAllConnections` need 18.2, so `>=18` could still hang (Codex P2)
- [x] T7  Make `startWebhook`'s shutdown idempotent: a `shuttingDown` guard schedules a single force-timer, so repeated signals can't leak a timer that fires after resolve (Codex)
- [x] T8  Fix the off-by-one wording in the conflict-retry comment (N retries, throw on conflict N+1) (Codex)
- [x] T9  Close test gaps: `startWebhook` signal-driven shutdown (idle close, force-close on timeout, repeated signals, handler cleanup) and `longPoll` conflict counter resets on a successful (incl. empty) poll (Codex)
- [x] T10  Simplify: trim the verbose polling/shutdown comments and JSDoc, collapse the conflict/plain log into one line (no behavior change)
- [x] T11  Refactor: move 409 classification (`isPollConflict` + `HTTP_STATUS_CONFLICT`) into `errors.ts` next to `isTransientError`, so all error taxonomy lives in one module
- [x] T12  Refactor: extract the shared SIGINT/SIGTERM install-run-cleanup into `src/node/withShutdownSignals`, used by both `run()` and `startWebhook()` (one place for the signal set)
- [x] T13  Refactor: extract `gracefulClose(server, timeoutMs)` in `server.ts` (close + drop idle + bounded force-close) so shutdown mechanics are self-contained and directly unit-testable
- [x] T14  Handle request aborts from a forced shutdown: `createWebhookServer` now `.catch`es the handler promise, so a socket destroyed mid-body (ECONNRESET from `readBody`) can't become an unhandled rejection that crashes the process mid-shutdown; regression test (Codex P2)
- [x] T15  Install `startWebhook`'s shutdown handlers only after the server is listening, so an early signal can't call `server.close()` on a not-yet-listening server (ERR_SERVER_NOT_RUNNING) - no try/catch needed (Copilot)
- [x] T16  `longPoll`: reset the conflict counter on a non-conflict transient error so the bound is on truly *consecutive* 409s (409 -> 5xx -> 409 no longer counts as two); test (Copilot)
- [x] T17  Stop mutating process-global signal listeners in the shutdown test: snapshot + diff to invoke startWebhook's own handler directly, avoiding flakiness under concurrent test runs (Copilot)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,7 +5,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 ## What this is
 
 `node-telegram-bot-api` v2: a from-scratch, runtime-agnostic TypeScript Telegram Bot API client.
-Node 18+, also runs on Bun, Deno, Cloudflare Workers, and Vercel/Deno Edge. The source is
+Node 18.2+, also runs on Bun, Deno, Cloudflare Workers, and Vercel/Deno Edge. The source is
 ESM/web-standard; the **published** package is dual-module - `zshy` emits both ESM (`*.js`/`*.d.ts`)
 and CJS (`*.cjs`/`*.d.cts`), exposed via the `exports` map's `import`/`require` conditions, so it can
 be `import`ed or `require`d. `src/core` stays Node-free, so the edge story is unchanged. There is
@@ -51,7 +51,7 @@ src/
           serializeParams + encodeForm, InputFile, the markup / entity / media builders
           (optional sugar), longPoll, webhookCallback, framework webhook adapters
           (Express / Next.js), errors. Web-standard APIs only.
-          -> zero node:* imports; runs on Node 18+ / Bun / Deno / Workers / edge.
+          -> zero node:* imports; runs on Node 18.2+ / Bun / Deno / Workers / edge.
   node/   the ONLY folder allowed to import node:*. Node-only sugar: fromPath() (fs
           uploads), createWebhookServer() (node:http -> delegates to core's
           webhookCallback), the managed run() polling runner, the DEBUG stderr sink.
--- a/doc/api.md
+++ b/doc/api.md
@@ -1212,7 +1212,9 @@ field name and attaches each part - it still stringifies nothing (ADR-011).
 | Property | Type |
 | --- | --- |
 | `allowedUpdates`? | string[] |
+| `conflictRetryDelayMs`? | number |
 | `limit`? | number |
+| `maxConflictRetries`? | number |
 | `offset`? | number |
 | `onError`? | (err: unknown) => void |
 | `retry`? | boolean |
--- a/doc/api.md
+++ b/doc/api.md
@@ -1079,6 +1079,13 @@ setup runs first (via `bot.startPolling` -> `bot.init()`), so a bad session
 store fails before the first poll; teardown (`bot.close()`) runs on the way
 out, whether the loop stopped or threw.
 
+A fatal poll-stop (the pump threw - a non-retriable error, or 409 conflicts
+past `maxConflictRetries`) is surfaced to stderr before it re-throws. As the
+managed lifecycle runner this is deliberate: a rejection alone can be dropped
+(fire-and-forget, or a `.catch` that swallows), and would then leave the
+process alive but no longer polling - the exact silent-hang #1350 describes.
+The error is still re-thrown unchanged, so an awaiting caller sees it too.
+
 | Param | Type |
 | --- | --- |
 | `bot` | [Bot](#bot) |
@@ -1107,6 +1114,12 @@ Managed webhook runner: create a `node:http` webhook server, start listening,
 and resolve when it shuts down. Installs `SIGINT`/`SIGTERM` handlers that close
 the server for a graceful exit (cleaned up in a `finally`), mirroring `run()` for
 long polling. Rejects if the server fails (e.g. the port is in use).
+
+Shutdown cannot hang: `server.close()` alone waits for every existing
+connection to end, so a single idle keep-alive socket would keep the `"close"`
+event (and this promise) pending forever - a stopped-but-never-exiting process.
+So we also drop idle connections immediately and force-close any still busy
+past `shutdownTimeoutMs`.
 
 You still register the webhook with Telegram yourself, pointing at this server's
 public URL (terminate TLS at a proxy/tunnel in front of it):
@@ -1291,6 +1304,7 @@ Options for a table block; `caption` is rich text.
 | `path`? | string |
 | `port` | number |
 | `secretToken`? | string |
+| `shutdownTimeoutMs`? | number |
 | `waitUntil`? | (promise: Promise<unknown>) => void |
 
 ### `TransportOptions`
--- a/doc/api.md
+++ b/doc/api.md
@@ -873,6 +873,21 @@ basename; pass `meta.filename` / `meta.contentType` to override.
 
 **Returns:** Promise<[InputFile](#inputfile)>
 
+### `gracefulClose()`
+
+Begin a non-hanging shutdown of `server`: stop accepting, drop idle keep-alive
+sockets at once (else `close` waits for them forever), and force-close anything
+still busy past `timeoutMs`. Returns the force-close timer so the caller can
+cancel it once `close` completes on its own. The connection helpers need Node
+18.2+ and are optional-chained so a non-Node runtime is a safe no-op.
+
+| Param | Type |
+| --- | --- |
+| `server` | Server |
+| `timeoutMs` | number |
+
+**Returns:** Timeout
+
 ### `isAbortError()`
 
 True for an `AbortController`/timeout abort, across runtimes. Matches both the
@@ -902,6 +917,18 @@ Cloudflare Workers), so the transport classifies our own client timeout as a
 | `value` | unknown |
 
 **Returns:** value is [InputFile](#inputfile)
+
+### `isPollConflict()`
+
+A 409 from `getUpdates` - another instance is polling the same token.
+Recoverable for polling (the competing poller usually exits), not for a plain
+request, so it is classified separately from `isTransientError`.
+
+| Param | Type |
+| --- | --- |
+| `err` | unknown |
+
+**Returns:** boolean
 
 ### `isTransientError()`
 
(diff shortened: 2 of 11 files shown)


yagop:

**Done**
- T1 (`702d821`): `longPoll` treats a 409 as recoverable *for polling* - it backs off with its own longer delay (`conflictRetryDelayMs`, default 5s) and resumes without advancing the offset, bounded by `maxConflictRetries` (default 10) consecutive conflicts before it throws. The streak resets on any successful poll; `onError` + a `log()` line make it observable. Truly fatal errors (401/400/404) still throw and stop the loop, unchanged.
- T2 (`23636ab`): unit tests - resume after a 409 then yield (offset not advanced), throw after N consecutive 409s (with onError count), and 409 fatal under `retry: false`.
- T3 (`fa1c6b3`): regenerated `doc/api.md` for the two new `LongPollOptions` fields.

**CI**: GitHub Actions (typecheck, build, unit on Bun + Node 22/24/26) and AppVeyor all pass; E2E skipped by design. Full local gate (`npm run check`, 225 tests) green.

Waiting for your merge. 🐾

yagop:
😺 Michi — added T4 for the second half of #1350 (the "process lingers dead" symptom); CI green.

**Investigation:** the core pump does *not* itself hang — verified empirically: `startPolling` rejects in ~4ms on a fatal poll error with `isRunning()` -> false, and `run()` removes its SIGINT/SIGTERM listeners + calls `bot.close()` in a finally, then re-throws (process exits once the rejection is observed). A process lingering *dead* after a fatal stop therefore requires an external open handle (a session-store pool, an HTTP server) keeping Node alive while polling is dead. Two fixes:

- **T1 fixes the actual incident**: a stray `getUpdates` (debug curl / overlapping deploy) returns 409, which no longer stops polling — it backs off and resumes, so nothing ever strands.
- **T4 (`77833cd`)** closes the residual for *any* fatal poll-stop (conflict bound exceeded, 401, ...): the managed `run()` runner now surfaces it to stderr before re-throwing, so a dropped or late-observed rejection can't leave the bot silently not-polling. The error is still re-thrown unchanged for awaiting callers. Unit test covers both the fatal (writes + re-throws) and clean-stop (silent) paths.

Note: preventing the *lingering* itself is inherently the caller's open resource (or T1) — a library can't force-close handles it doesn't own — so T4 makes the failure unmissable rather than force-killing the process. Happy to instead add an `onStop(err)` hook or an opt-in exit-on-fatal if you'd prefer a different policy.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 227 tests) + Node 22/24/26 unit run green.

Waiting for your merge. 🐾

yagop:
😺 Michi — added T5, the real "not stopping" bug, found by tracing the mitubo-bot commits; CI green.

The production hang wasn't the poll loop — it was **webhook-server shutdown** hanging on a stuck connection during Fly scale-to-zero idle-exit (fixed downstream in mitubo-bot `1a09e2a2` with a `server.stop()` timeout). ntba's own managed `startWebhook` (`src/node/server.ts`) had the **identical gap**: `node:http`'s `server.close()` waits for every existing connection to end, so one idle keep-alive socket keeps the `"close"` event — and the `startWebhook` promise — pending forever. On SIGTERM the process stops serving but never exits: "no work, but not stopping."

**T5 (`ed16561`):** shutdown now drops idle connections immediately (`closeIdleConnections()`) and force-closes anything still busy past `shutdownTimeoutMs` (new option, default 10s, mirroring mitubo-bot) via `closeAllConnections()`, both optional-chained for cross-runtime safety. Tests: `close()` completes despite an idle keep-alive socket, and `startWebhook` rejects + cleans up its signal handlers on a listen error. Docs regenerated for the new option.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 229 tests) + Node 22/24/26 unit run green.

Waiting for your merge. 🐾

yagop:
😺 Michi — applied the Codex (gpt-6-astra) review; CI green.

- **[P2] Shutdown could still hang on Node 18.0/18.1** (`closeIdleConnections`/`closeAllConnections` are 18.2+): **T6 (`ab28496`)** bumps the documented minimum to Node 18.2 (`engines: >=18.2.0`, CLAUDE.md), so the helpers are guaranteed on Node; the optional-chaining stays only as a safe no-op for non-Node runtimes, now noted in the comment.
- **Repeated signals leave a timer behind**: **T7 (`8b16bba`)** makes `stop()` idempotent via a `shuttingDown` guard, so a second signal can't schedule a second force-timer that outlives resolve.
- **Off-by-one in the conflict-retry comment**: **T8 (`35ecef0`)** reworded to "retry on up to N consecutive conflicts; the next one throws."
- **Test-coverage gaps**: **T9 (`b462eca`)** adds a `longPoll` test that a successful (empty) poll resets the conflict counter, and a `startWebhook` test that drives the real signal path — installs the handler, drops an idle keep-alive connection on SIGTERM, is idempotent under a repeated signal, and cleans up its listeners.

One deliberate scope note: I did not add an explicit force-`closeAllConnections`-on-timeout test — it needs a reliably-stuck (never-completing) socket, which is flaky to construct deterministically. The timeout path is exercised (the timer is created/cleared) and the force-close is a single guarded call; happy to add a child-process integration test if you'd prefer belt-and-suspenders.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 231 tests) + Node 22/24/26 + typecheck:test green.

Waiting for your merge. 🐾

yagop:
😺 Michi — applied the Copilot review; CI green.

- **[server.ts] `gracefulClose` could throw `ERR_SERVER_NOT_RUNNING`** if a signal arrived before `listen()` finished. **T15 (`bac47f9`)** fixes it structurally rather than with a try/catch: `startWebhook` now listens first (phase 1) and only installs the `SIGINT`/`SIGTERM` handlers once the server is actually listening (phase 2). So `stop()` can never run `server.close()` on a not-yet-listening server, and `gracefulClose` stays guard-free. A signal during the brief startup window keeps Node's default (exit), which is correct - nothing is listening to close.
- **[longpoll.ts] conflict bound wasn't truly "consecutive"**: a non-conflict transient between two 409s (409 -> 5xx -> 409) still accumulated. **T16 (`3a4783c`)** resets the counter on any non-conflict transient, so only genuinely consecutive 409s count toward `maxConflictRetries`; added a test (409 -> 500 -> 409 with `maxConflictRetries: 1` no longer throws).
- **[test] global signal-listener mutation**: **T17 (`5fcc057`)** stops `removeAllListeners`/`emit` on the process; the test now snapshots the SIGTERM listeners, lets `startWebhook` add its own, and invokes *that* handler directly - no global state touched, so it can't disturb other tests or the runner.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 233 tests) + Node 22/24/26 + typecheck:test green.

Waiting for your merge. 🐾

PR #1360, opened 2026-09-06
run(): a failed polling pump leaves the process alive and silent (#1351)

### LINKED ISSUE #1351: run(): a failed polling pump leaves the process alive and silent
## What happens

When the polling pump fails, `run()` (`src/node/run.ts`) rejects - and a process whose entry point is the documented

```ts
await run(bot);
```

keeps running with **no output at all**. It looks alive to every outside observer (`ps` shows it, no exit code, no log line), while it polls nothing and handles nothing.

That is the worst failure shape available: a crash would page someone, a log line would explain it, but a silent live process just looks like a bot that stopped answering.

## Repro

Take any example ending in `await run(bot)` (e.g. `examples/17-callback-tracking.ts`), start it, then force the pump to fail - a `409 Conflict` is the easy way (see #1350):

```sh
curl -s "https://api.telegram.org/bot$TOKEN/getUpdates?timeout=0&limit=1"
```

Observed: the process stays up, no `getUpdates` goes out again, stderr stays empty. Only `DEBUG='*'` (which stops at the last successful cycle) hints that polling ended.

## Why it is silent

`run()` returns `bot.startPolling(...)`; the rejection propagates to the caller's top-level `await`, and how *that* surfaces is up to the runtime and whether anything else keeps the event loop alive. In practice the process neither prints nor exits reliably.

## Suggested fix

`run()` is the managed runner - the place that already owns signal handlers and teardown - so it should own failure visibility too. Options, roughly in order of preference:

1. Log the failure before rethrowing (`console.error`, matching `Bot`'s default error boundary), so the reason is always on stderr even if the rejection goes nowhere.
2. Optionally exit the process with a non-zero code (`opts.exitOnError`, default on?) so a supervisor restarts the bot - which is what an operator expects from a managed runner.
3. At minimum, document that `run()` rejects and that callers must handle it:
   ```ts
   run(bot).catch((err) => { console.error(err); process.exit(1); });
   ```
   and update the examples, which all use a bare `await run(bot)`.

Whatever we pick, the examples should model it, since they are what people copy.

## Notes

Found while running `examples/17-callback-tracking.ts` live against a real token; the bot had stopped polling minutes earlier with nothing to show for it. Related: #1350 (what made the pump fail in the first place).

yagop:
Closes #1351

Suggestion #1 (log the fatal poll-stop to stderr before rethrowing) already landed in #1358. This PR adds the remaining scope.

- [x] T1  `run()`: add opt-in `exitOnError` (default false) - on a fatal poll-stop, after teardown + the stderr log, exit non-zero so a supervisor restarts the bot; new `RunOptions` type; unit test (stubbed `process.exit`)
- [x] T2  Update the examples that end in a bare `await run(bot)` to model failure handling (`await run(bot, { exitOnError: true })`)
- [x] T3  Regenerate `doc/api.md` for the new `RunOptions` / `run` signature
- [x] T4  `RedisSessionStorage`: add a `url` option so the store can own the client it opens and `close()` it on teardown (mirrors the sql/sqlite `owned` pattern) - closes the one teardown gap found while verifying `bot.close()` for #1351; unit test + docs
- [x] T5  Harden `exitOnError` (Codex + Copilot): gate the hard exit on owning the pump (a losing double-run must not kill the healthy process), run it in a nested `finally` so a throwing teardown can't skip it, and flush the stderr diagnostic before exiting; tests for teardown-throws and lost-race
- [x] T6  Cover the url-owned Redis client path via a `protected createClient` seam - construct/close the owned client, idempotent close, and use-after-close rejects (Copilot)

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -437,6 +437,7 @@ bound for long-lived bots that talk to many distinct chats.
 
 | Method | Params | Returns | Description |
 | --- | --- | --- | --- |
+| `close` | - | void | Close the client, but only if this store opened it (a passed-in `client` and the shared `redis` are the caller's / runtime's). Closing ends this store's life - a later use throws and you construct a new one. With a shared or caller-supplied client this is a no-op. |
 | `delete` | `key`: string | Promise<void> | - |
 | `read` | `key`: string | Promise<string \| undefined> | The stored string for `key`, or `undefined` when there is none. |
 | `touch` | `key`: string, `ttlSeconds`: number | Promise<void> | Refresh a key's expiry without rewriting it - what the middleware calls when an update changed nothing, so an active chat is not evicted mid-conversation. |
@@ -6346,6 +6347,7 @@ type RedisSessionStorageOptions = {
   client?: RedisClient;
   prefix?: string;
   ttlSeconds?: number;
+  url?: string;
 };
 ```
 
--- a/doc/api.md
+++ b/doc/api.md
@@ -1109,12 +1109,13 @@ out, whether the loop stopped or threw.
 A fatal poll-stop is also written to stderr before being re-thrown: a bare
 rejection can be dropped (fire-and-forget, or a swallowing `.catch`), which
 would leave the process alive but no longer polling - the silent hang #1350
-describes.
+describes. Pass `exitOnError: true` to also exit the process non-zero after
+teardown, so a supervisor restarts the bot instead of relying on the caller.
 
 | Param | Type |
 | --- | --- |
 | `bot` | [Bot](#bot) |
-| `options?` | [LongPollOptions](#longpolloptions) |
+| `options` | [RunOptions](#runoptions) |
 
 **Returns:** Promise<void>
 
@@ -1136,9 +1137,10 @@ comparison always inspects every position, so it leaks no information about
 ### `startWebhook()`
 
 Managed webhook runner: create a `node:http` webhook server, start listening,
-and resolve when it shuts down. Installs `SIGINT`/`SIGTERM` handlers that close
-the server for a graceful exit (cleaned up in a `finally`), mirroring `run()` for
-long polling. Rejects if the server fails (e.g. the port is in use).
+and resolve when it shuts down. Once listening, installs `SIGINT`/`SIGTERM`
+handlers that close the server for a graceful exit (cleaned up in a `finally`),
+mirroring `run()` for long polling. Rejects if the server fails to start (e.g.
+the port is in use).
 
 Shutdown cannot hang: `server.close()` waits for existing connections to end,
 so we also drop idle keep-alive sockets at once and force-close anything still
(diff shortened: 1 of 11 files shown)


yagop:

Suggestion #1 (log the fatal poll-stop to stderr before rethrowing) landed earlier in #1358, so this PR covers the rest:

- **T1 (`e6865ba`)**: `run()` gains an opt-in `exitOnError` (new `RunOptions` type). On a fatal poll-stop it still logs to stderr and re-throws, and *additionally* - after teardown (`bot.close()`) - exits the process non-zero, so a supervisor (systemd/Docker/pm2) restarts a clean process instead of a live-but-silent one. Default `false` to preserve the re-throw contract and avoid a surprise `process.exit` (which would also kill test runners). Unit tests stub `process.exit` and cover the fatal-exit path and no-exit-on-clean-stop.
- **T2 (`7015c20`)**: the six examples that ended in a bare `await run(bot)` now use `await run(bot, { exitOnError: true })`, with a comment on the primary example explaining the supervisor-restart rationale (they're what people copy).
- **T3 (`f8b31ea`)**: regenerated `doc/api.md` for `RunOptions` / the `run` signature (also swept in a stale `startWebhook` docstring line #1358's regen had missed).

Note on the default: I chose `exitOnError: false` rather than the issue's tentative "default on?" - a library function that calls `process.exit` by default is surprising and would terminate test/embedding processes. The examples opt in, which is where the supervisor pattern belongs. Easy to flip to default-on if you'd prefer.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 237 tests) + typecheck:examples green.

Waiting for your merge. 🐾

yagop:
😺 Michi — applied the Codex + Copilot review; CI green.

- **Exit skipped if teardown throws** (Codex + Copilot): **T5 (`71cb9fb`)** moves the hard exit into a nested `finally`, so `exitOnError` fires even when `bot.close()` rejects - it can't be defeated by the very stuck resource it exists to escape.
- **Exit fired when this call doesn't own the pump** (Copilot): T5 gates the exit on `owned` (via a `failed` flag only set on the owning path). A second `run(bot, { exitOnError: true })` that loses the already-running race now just re-throws to its caller instead of `process.exit`-ing the healthy pump's process. The misleading "polling stopped" stderr line is also gated on ownership.
- **stderr truncated by process.exit** (Copilot): T5 awaits the `process.stderr.write` flush callback before exiting, so the fatal diagnostic survives even when stderr is piped to a supervisor. Test stub updated to invoke the callback.
- **Redis owned-`url` path untested** (Copilot): **T6 (`02a8366`)** adds a `protected createClient()` seam (excluded from public docs) so a fake owned client is testable without a live Redis; new test covers construct -> close (owned) -> idempotent -> use-after-close rejects. Added an `open()` guard so a reused closed owned store throws a clear "was closed" error (parity with the sql/sqlite stores).

New tests: teardown-throws-still-exits, lost-race-does-not-exit, and the full url-owned Redis lifecycle.

**CI**: all GitHub Actions + AppVeyor green; E2E skipped by design. Full local gate (`npm run check`, 241 tests) + typecheck:bun/test/examples green.

Waiting for your merge. 🐾

yagop:
😺 Michi — merged, all done 🐾

